PR #1398, opened 2015-01-08
Supplyorder items restructured to attribute on SupplyOrder. (was archety...

anneline:
...pe)


xispa:
SupplyOrder stuff is not my circus, not my monkeys, but the .po files modifications shouldn't be there I guess. The rest, seems pretty ok to me. A test-case should be nice too.

PR #1399, opened 2015-01-08
HEALTH-203 KeyError: 'searchTerm'

Espurna:
http://sentry.bikalabs.com/bikalims/healthbikalabscom/group/5026/

Preventive improvements.


xispa:
Not sure if ajaxGetClients is still used, but accepted for back-compatibility with health's extension

PR #1400, opened 2015-01-12
Some staff from the issue resolved on Health-179

Espurna:


xispa:
A silent error rises when Bika-LIMS' JSON-API tries to access to ComputedFields declared in a content type from a child extension, but only if the expression includes a direct access (not using the Schema()) to a method or property. This fix prevents the silent error.

PR #1413, opened 2015-01-20
Feature/related issue health 200

Espurna:
Quick edit button implemented. It works like addbutton.

No test is nedded now, since it is implemented for health.


xispa:
get_ovelaybutton_options doesn't take into account the add_button.

PR #1428, opened 2015-01-30
shashattr instead of hasattr - avoid acquisition.

rockfruit:
One of my test sites is called 'asdf'.  Then, I made a service with
keyword 'asdf'.  Suddenly this code thought that my site object was
a service, since 'asdf' is an attribute of the root.

### DIFF
--- a/bika/lims/browser/fields/aranalysesfield.py
+++ b/bika/lims/browser/fields/aranalysesfield.py
@@ -146,7 +146,7 @@ class ARAnalysesField(ObjectField):
                 interim_fields.append(v)
 
             # create the analysis if it doesn't exist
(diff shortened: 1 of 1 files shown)


xispa:
mmm.. didn't know about [shasattr](http://apidoc.eea.europa.eu/Products.Archetypes.utils-pysrc.html#L695). Specially intersting the comment below:

```
It should not swallow all errors, especially now that descriptors make computed
attributes quite common.  getattr() only recently started catching only AttributeErrors,
but apparently hasattr is lagging behind.  I suggest the consistency between getattr
and hasattr should be fixed in Python, not Zope.

PR #1429, opened 2015-02-02
LIMS-257 Set Blank and Warning icons in Reference Sample main view

Espurna:


bika/lims/browser/templates/referencesample_view.pt
```diff
@@ -12,6 +12,17 @@
         <img tal:condition="view/icon | nothing"
             src="" tal:attributes="src view/icon"/>
         <span tal:content="context/title_or_id"/>
+        <span tal:define="portal context/@@plone_portal_state/portal;">
```
xispa: I wouldn't place this span here just only to retrieve the portal object. Maybe you can add the tal:define in &lt;body&gt; and remove this span. Then the 'portal' object will be accessible from everywhere inside the template...
```html
<body tal:define="portal context/@@plone_portal_state/portal;">
```
and you can then remove the 'portal' var declared inside metal:content element (#L32)

PR #1431, opened 2015-02-02
date in header table saves and renders correctly

rockfruit:
Needs a spanish eye to look it over :)

This seems to fix some other problem, where dates would appear in different formats in different places, regardless of settings.

### DIFF
--- a/bika/lims/browser/__init__.py
+++ b/bika/lims/browser/__init__.py
@@ -1,17 +1,62 @@
 """Bika's browser views are based on this one, for a nice set of utilities.
 """
+from DateTime.DateTime import DateTime, safelocaltime
+from DateTime.interfaces import DateTimeError
 from Products.CMFCore.utils import getToolByName
 from AccessControl import ClassSecurityInfo
-from Products.CMFPlone.i18nl10n import ulocalized_time
+from Products.CMFPlone.i18nl10n import ulocalized_time as _ut
 from Products.Five.browser import BrowserView
 from bika.lims import logger
 from zope.cachedescriptors.property import Lazy as lazy_property
 from zope.i18n import translate
-import plone, json
+from time import strptime as _strptime
+
+
+def strptime(context, value):
+    """given a string, this function tries to return a DateTime.DateTime object
+    with the date formats from i18n translations
+    """
+    val = ""
+    for fmt in ['date_format_long', 'date_format_short']:
+        fmtstr = context.translate(fmt, domain='bika', mapping={})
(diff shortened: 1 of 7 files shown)


bika/lims/browser/__init__.py
```diff
@@ -1,17 +1,62 @@
...
+        fmtstr = fmtstr.replace(r"${", '%').replace('}', '')
+        try:
+            val = _strptime(value, fmtstr)
+        except ValueError:
+            continue
+        try:
+            val = DateTime(*list(val)[:-6])
+        except DateTimeError:
+            val = ""
+        if val.timezoneNaive():
```
xispa: To prevent an unhandled error, better if:
```python
if val and val.timezoneNaive():
```
or much more better if
```python
try:
    val = DateTime(*list(val)[:-6])
except DateTimeError:
    continue
```

bika/lims/browser/__init__.py
```diff
@@ -1,17 +1,62 @@
...
+            # Use local timezone for tz naive strings
+            # see http://dev.plone.org/plone/ticket/10141
+            zone = val.localZone(safelocaltime(val.timeTime()))
+            parts = val.parts()[:-1] + (zone,)
+            val = DateTime(*parts)
+        break
+    else:
+        logger.warning("DateTimeField failed to format date "
+                       "string '%s' with '%s'" % (value, fmtstr))
```
xispa: Since there are those 2 try/catch with "continue" inside the loop, I wouldn't anchor this "else" to the loop. I'd remove that else and use an if instead:
```python
if not val:
    logger.warning("DateTimeField failed to format date "
	               "string '%s' with '%s'" % (value, fmtstr))
return val
```
This way, we'll prevent us from silent errors/warnings....

bika/lims/browser/fields/datetimefield.py
```diff
@@ -0,0 +1,70 @@
...
+                    # Use local timezone for tz naive strings
+                    # see http://dev.plone.org/plone/ticket/10141
+                    zone = val.localZone(safelocaltime(val.timeTime()))
+                    parts = val.parts()[:-1] + (zone,)
+                    val = DateTime(*parts)
+                break
+            else:
+                logger.warning("DateTimeField failed to format date "
+                               "string '%s' with '%s'" % (value, fmtstr))
+
```
xispa: All this code (from L44 to L64) maybe could be replaced by a direct call to bika.lims.browser:
```python
val = bika.lims.browser.strptime(instance, value)
```

PR #1433, opened 2015-02-05
Feature/lims 1539

xispa:
WS print view. In both AR by row or column orientations


xispa:
Background color for orfan analyses don't get printed...

PR #1434, opened 2015-02-06
Feature/lims 1539

xispa:
WS print view. In both AR by row or column orientations.
- WS ID, AR IDs and QC Sample IDs barcoded
- Orphan analyses' background in grey (in a WS, not all analyses are assigned to all ARs/QCs)
- Preview that allows to set the template and/or columns to be displayed dynamically
- Followed similar procedure as AR reports, so the user can create additional templates either by creating an add-on or by placing a tpl and a css file inside browser/worksheet/templates/print.
- The system is ready for multiple-WS print at once, but requires to add a "Print" button in WS folder (with this PR, there's only a "Print" button inside WS/manage view.


xispa:
@rockfruit I think you did a partial merge.. maybe you forgot to do a pull from the latest xispa:feature/LIMS-1539
https://github.com/xispa/Bika-LIMS/compare/feature/LIMS-1539

rockfruit:
Thanks!  That should do it.

PR #1438, opened 2015-02-10
Added option to select landing page for clients in configuration registry

zylinx:
review only, still need to write tests


zylinx:
It's ready now.

PR #1485, opened 2015-04-20
LIMS-1745 Retracted analyses in duplicates

jayadeepk:
In case there are retracted analyses, src_analyses gets three analyses for a given 'analysis_uid' and hence three duplicates. To avoid this the loop on src_analysis run only once. Though might be a temporary fix, it doesn't affect other code much.


rockfruit:
neat!

PR #1488, opened 2015-04-22
LIMS-1748 Fixed error in adding supply order

jayadeepk:
The supply order gives the error when there is a product with no price given. Price is not a mandatory field for products. The fix allows the supply order to neglect such products because price is required when supply order is created. [LIMS-1748](https://jira.bikalabs.com/browse/LIMS-1748).

### DIFF
--- a/bika/lims/browser/supplyorder.py
+++ b/bika/lims/browser/supplyorder.py
@@ -110,6 +110,8 @@ class EditView(BrowserView):
             self.products = []
             products = sorted(products, key = methodcaller('Title'))
             for product in products:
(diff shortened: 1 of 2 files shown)


xispa:
I'd rather prefer the price to be a mandatory field for products. If they want to set a price, use 0.

jayadeepk:
@xispa : Yeah, it'd be great too. I'll make another pull request for it. Guess I can close this now.

xispa:
Thanks @jayadeepk ! merge closed!.

PR #1491, opened 2015-04-23
LIMS-1741 Fixed overlay when saving supply order

jayadeepk:
The unwanted overlay pops up because the form has `name="edit"`. I traced back where the overlay came from and it is `plone.app.discussion.javascript/comments.js`. It added prepOverlay to all forms with `name="edit"` for some unwanted reason.

Removing `name="edit"` not only fixed the issue, but also gave no other errors in the supply order workflow.


anneline:
Cool! thanks!

PR #1492, opened 2015-04-27
LIMS-1754 Easy install for LIMS' add-ons was not possible

Espurna:


xispa:
You need to add the 'z3c.unconfigure' dependency in the setup and/or buildout.cfg

PR #1504, opened 2015-05-14
LIMS-280 System IDs starting from a specific value

jayadeepk:
The specific value can be configured using Sequence Start in Bika Setup. The sequence_start applies only if it is greater than existing IDs. Sample is treated as special case with new setup field called SampleIDSequenceStart. In bika setup, types of padding and sequence_start are made 'int' to provide validation.

Adding samples, ARImports, batches, clients and worksheets has been thoroughly tested manually (with and without) specifying sequence_start.

### DIFF
--- a/bika/lims/content/bikasetup.py
+++ b/bika/lims/content/bikasetup.py
@@ -23,21 +23,26 @@ class PrefixesField(RecordsField):
     _properties = RecordsField._properties.copy()
     _properties.update({
         'type' : 'prefixes',
-        'subfields' : ('portal_type', 'prefix', 'separator', 'padding'),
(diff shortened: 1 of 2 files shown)


jayadeepk:
Excuse me, there was some error that repeats an ID. I'll fix it and send a new one.

rockfruit:
@jayadeepk we're trying to take care to add functional/unit tests for new code which is committed to hotfix/ branches.  I know it's a royal pain, but have a look into bika/lims/tests/test_*.py and see if there's any way you could force a test for this.  It should be quite simple, but I'm learning the same techniques right now; pop into IRC or send me a mail if you get stuck for more than 10 minutes on this;  Wutwut, woop!

PR #1512, opened 2015-05-20
Windows peculiarities

anneline:
and syntax errors (which only Windows - and I - care about)


rockfruit:
Good to have a windows fiend onboard... always remember to keep a few of those around.

PR #1518, opened 2015-05-21
LIMS-280 System IDs starting from a specific value

jayadeepk:
The specific value can be configured using 'Sequence Start' in Bika Setup. It applies only if it is greater than existing IDs.
- Sample is treated as a special case
- In bika setup, types of padding and sequence_start  are made 'int' for validation.
- Added tests for checking IDs of samples and batches after changing sequence_start.


jayadeepk:
Fixed merge conflicts. @rockfruit

jayadeepk:
The merge conflict is only in CHANGELOG and keeps popping as other pull requests get accepted. Resolving it again.

rockfruit:
Thanks! perfect-o-mondo.

PR #1531, opened 2015-05-27
LIMS-1697 Error updateing bika.lims 317 to 318 via quickinstaller

xispa:
Yes... I know a massive and ugly hack, but fix the upgrade issues rised because of the chestnut `source="*"` mentioned here: https://github.com/bikalabs/Bika-LIMS/pull/1484


xispa:
Needs more testing

PR #1536, opened 2015-05-28
Tabs replacement by whitespaces

xispa:
Haven't done any change in the logic


rockfruit:
@xispa thank you, for taking the time to pay attention at this level of detail.

PR #1548, opened 2015-06-10
LIMS-1695 Invoice export function missing?

Espurna:
An upgrade step is needed.
@lemoene I don't know if this esporter generates the csv format that you want... This is an examples:

```
Invoice Batch
ID,invoicebatch.2015-06-03.8239692103
Invoice Batch Title,Jun 2015 - ad hoc
Start Date,2015-06-01
End Date,2015-06-30

Invoices
Invoice ID,Client ID,Client Name,Account Num.,Phone,Date,Total Price
I-001,client-1,Happy Hills,,021 555 4220,2015-06-03

Invoices items
Date,Description,Order,Amount,VAT,Amount incl. VAT
,Barley ,BAR-0001-R01,10.00,1.4,11.4
```

Issue: https://jira.bikalabs.com/browse/LIMS-1695


lemoene:
Thanks Pau, looks right. I would not recognise the Pastel format immediately but it if it is anything like Bika 2's we are safe for now. We'll find out when acceptance testing.

Copying your comment to the issue too

lemoene:
Thanks Pau, looks right

PR #1549, opened 2015-06-10
LIMS-1864 Product Item entity

Ammy2:
Added productitem content type via ProductItem and ProductItems, under bika_setup.


xispa:
Cool!. Remember to set the JIRA's ticket ID in the title!

PR #1555, opened 2015-06-16
Fix regression importing navigation portlet (ConstraintNotSatisfied, …

rockfruit:
…Error Value: /)

https://dev.plone.org/ticket/10232


rockfruit:
This error prevents site from being created with Plone 4.3.6

PR #1558, opened 2015-06-18
Overrode breadcrumbs in Product view

jayadeepk:
Clicking on a supplier in the breadcrumbs of Product view redirects to Products list instead of reference samples.


jayadeepk:
Just updated the branch by removing some empty lines in the Viewlet.

PR #1559, opened 2015-06-18
Default arreport template

rockfruit:


rockfruit:
Should also fix LIMS-1886

xispa:
Cool!. I didn't notice LIMS-1886 before. To me, only in windows machines... I am working a lot now with results reports and I haven't experienced that..

PR #1560, opened 2015-06-18
LIMS-1887 uniquefield validator doesn't work properly

Espurna:
https://jira.bikalabs.com/browse/LIMS-1887

A pdb trace example:

-> for item in aq_parent(instance).objectValues():
(Pdb) item
<Method at /Plone/methods/method-1>
(Pdb) hasattr(item, 'UID') and item.UID() != instance.UID() and fieldname in item.Schema() and item.Schema()[fieldname].get(item) == value
False
(Pdb) value
'1212'
(Pdb) item.Schema()[fieldname].get(it)
1212
(Pdb) type(value) == type(it.Schema()[fieldname].get(it))
False
(Pdb) type(it.Schema()[fieldname].get(it))
<type 'int'>

### DIFF
--- a/bika/lims/validators.py
+++ b/bika/lims/validators.py
@@ -33,7 +33,13 @@ class UniqueFieldValidator:
         for item in aq_parent(instance).objectValues():
             if hasattr(item, 'UID') and item.UID() != instance.UID() and \
                fieldname in item.Schema() and \
-               item.Schema()[fieldname].get(item) == value:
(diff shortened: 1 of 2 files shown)


rockfruit:
thanks.

May be unrelated, but... It seems that having a "StringField" affects the widgets presented, and these widgets will generally insert a String value into the field, but if you pass the setter an int, dict or other object type, it will be stored verbatim.

PR #1565, opened 2015-06-23
LIMS-1900 Add Robot tests for creating Orders

Ammy2:


rockfruit:
I think all our specs should be written in this beautiful form of english.

PR #1585, opened 2015-07-21
LIMS-1941 Error on manage_results view due to Instruments validity check

xispa:


bika/lims/browser/analysisrequest/manage_results.py
```diff
@@ -74,8 +75,17 @@ def checkInstrumentsValidity(self):
             will be displayed.
         """
         invalid = []
-        ans = [a.getObject() for a in self.context.getAnalyses()]
+        ans = self.context.getAnalyses(full_objects=True)
         for an in ans:
+            try:
+                an = an.getObject()
```
rockfruit: this will always fail, since getObject never exists on an analysis instance.  So this for-loop should have no effect, except to print traceback for every single analysis. Do have data, or do you think labsanmartin can send some data for reproducing the error?

I remember discussing it with leo in irc, and I would LOVE to find it.

rockfruit: It seems like something that is not a routine Analysis has been inserted into the AR.

xispa: I see, you're right, this code is ridiculous and hides the "real" problem. We only have the error stack, seems that one of the elements returned from getAnalyses() is an string instead of an Analysis object.

```
Hora
11/07/2015 11:18
        Nombre del Usuario
xxxx (xxxx)
        URL de la Petición
http://xxx.xxx.xxx.xxx:8085/Plone/clients/client-19/AA15-1557-R01/manage_results
        Tipo de Excepción
AttributeError
        Valor de la Excepción
21A051
        Traceback (innermost last):

Module ZPublisher.Publish, line 138, in publish
      * Module ZPublisher.mapply, line 77, in mapply
      * Module ZPublisher.Publish, line 48, in call_object
      * Module bika.lims.browser.analysisrequest.manage_results, line 64, in __call__
      * Module bika.lims.browser.analysisrequest.manage_results, line 77, in checkInstrumentsValidity
      * Module Products.ZCatalog.CatalogBrains, line 102, in getObject
      * Module OFS.Traversable, line 317, in restrictedTraverse
      * Module OFS.Traversable, line 285, in unrestrictedTraverse
      * __traceback_info__: ([], '21A051')
        
AttributeError: 21A051
```

leonardorojass:
I think this problem was also related to the AR which had 3 times the same AS. 21A051 is the keyword for that AS. This problem has not happened again.

PR #1609, opened 2015-08-26
LIMS-1841 Expiry alert for product item's cancellation

Ammy2:
A portlet to alert expiry of product-items which will expire within ~2 months time.

To check just add the Expiry Portlet under Manage Portlets and one can see the list of product-items which are going to expire within two months, along with their expiry date.

While adding the portlet one can specify the limit (default:5) of how many items should be shown in the list.


rockfruit:
TODO port existing 'classic' portlets to the new style

PR #1613, opened 2015-09-02
LIMS-1397 During site import, Client Title accessor fails.

rockfruit:
The code at git@github.com:bikalabs/bika-export-import.git fixes LIMS-1397 and other issues by generalising the import/export of setupdata.  However the code has yet to be tested and integrated into bika.lims.

During this import, the existing client accessor causes cataloging to fail.  This one is identical, but works.

### DIFF
--- a/bika/lims/content/client.py
+++ b/bika/lims/content/client.py
@@ -131,7 +131,8 @@ class Client(Organisation):
 
     def Title(self):
         """ Return the Organisation's Name as its title """
-        return safe_unicode(self.getField('Name').get(self)).encode('utf-8')
+        name = self.schema['Name'].get(self)
+        return safe_unicode(name).encode('utf-8')
 
     security.declarePublic('getContactFromUsername')
     def getContactFromUsername(self, username):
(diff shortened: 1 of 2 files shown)


xispa:
Woooaaa... bika-export-import.git sucks. Haven't gone in deep details yet, but maybe we need to figure out how to generalise the importer a little bit more to be used transparently by add-ons (i.e. Health). Maybe retrieving the 'export_types' directly from a file from inside the zip instead of 'hardcoding' them?
ref: https://github.com/bikalabs/bika-export-import/blob/master/import_bika_setup.py#L36

xispa:
Changelog? ;)

rockfruit:
Related to LIMS-1397, but does not fix it yet.  bika-import-export.git is temporary.  Code was written quickly for client use, and for minimal impact to existing configuration I kept it separate... It needs some testing with different configurations.  When it's been tested by more than just me, on more than one site (and probably after some tweaking of the formats, and adding Data Validation rules) we can include it in bika.lims

We could just apply a new marker interface to classes who's instances must be exported/imported with this machine. That allows the code to enumerate them, and a GUI can be applied to select which ones should be included in any particular export/import. Very simple: but this commit is just fixing the LIMS code so the bika-export-import stuff works.

rockfruit:
The main concern for Data Validation lookups, is that some services are permitted to have duplicate titles.  This allows a "Calcium" service to exist in a Metals and Water categories with the same name, while being a different service.  I'm still not sure that's a great idea, but it's useful to many folks.

PR #1622, opened 2015-09-16
LIMS-1545 Add "Composite Y/N" checkbox on AR Template

Espurna:
https://jira.bikalabs.com/browse/LIMS-1545

- Blocking filled SampleTypes fields in the ARAdds which comes from a SamplingRound
- Introducing the logic for composite selection in ARAdd when coming from a SamplingRound
- Composite field in ARTemplate
- Composite column in SRTemplate's ARtemplates listing view

### DIFF
--- a/bika/lims/browser/srtemplate/artemplates.py
+++ b/bika/lims/browser/srtemplate/artemplates.py
@@ -175,10 +175,10 @@ class AnalysisRequestTemplatesView(BikaListingView):
                     (obj.getSampleType().absolute_url(), item['SampleType'])
             else:
                 item['SampleType'] = ''
-            if obj.getComposite():
-                item['Composite'] = 'Y'
-            else:
-                item['Composite'] = 'N'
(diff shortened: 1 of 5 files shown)


bika/lims/browser/srtemplate/artemplates.py
```diff
@@ -175,6 +175,10 @@ def folderitems(self):
                     (obj.getSampleType().absolute_url(), item['SampleType'])
             else:
                 item['SampleType'] = ''
+            if obj.getComposite():
+                item['Composite'] = 'Y'
+            else:
+                item['Composite'] = 'N'
```
xispa: Better to show an icon if the sample is Composite instead of Y and leave the field blank if not composite, as we do in methods list for 'Manual Entry of Results':
https://github.com/bikalabs/Bika-LIMS/blob/hotfix/3.1.9/bika/lims/content/methods.py#L137

PR #1623, opened 2015-09-16
LIMS-1564 Cancelling a Sampling Round

Espurna:
https://jira.bikalabs.com/browse/LIMS-1564

- After canceling a sampling rounds, all its analysis request should be canceled too.


xispa:
From [LIMS-1564](https://jira.bikalabs.com/browse/LIMS-1564):
> NB. When the round is cancelled, all its associated **Samples** and ARs are cancelled by the system.

Hehe... you need to cancel the Sample too....

PR #1637, opened 2015-09-30
LIMS-1970 Analyses from AR Add form not displayed properly

Espurna:
JIRA: https://jira.bikalabs.com/browse/LIMS-1970

For some strange happenings when updating from 318, jquery's "replaceWith()" function was removing all TR's from the html variable 'data'


xispa:
aaah... relief

PR #1651, opened 2015-10-15
Dashboard. Stacked-bar charts instead of multi-bar charts

xispa:
Maybe this should go to develop, but I think is almost a cosmetic change and I think multi-bar charts in dashboard don't make sense.

stacked


rockfruit:
agree, there is only 0.001% chance that any problems will come from these kinds of changes....

really pretty!

rockfruit:
Also I believe that if we allow ourselves to blur the lines sometimes, it will keep us happier.

xispa:
Totally agree... by the way, please note I added some labels to better manage pull requests. Sometimes, when I need to fix something or code something new, comes in my mind a contribution done earlier that could be the cause (of the bug) or could help on the development of new code. The addition of labels and self-explanatory descriptions are really helpful to find those chunks of code.

PR #1660, opened 2015-10-24
Fix LIMS-2080: Correctly interpret default values

rockfruit:
Ran 2 tests with 0 failures and 0 errors in 11.131 seconds.

### DIFF
--- a/bika/lims/content/arimport.py
+++ b/bika/lims/content/arimport.py
@@ -537,18 +537,22 @@ class ARImport(BaseFolder):
 
             # ContainerType - not part of sample or AR schema
             if 'ContainerType' in row:
-                obj = self.lookup(('ContainerType',),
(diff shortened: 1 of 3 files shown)


bika/lims/content/arimport.py
```diff
@@ -537,18 +537,22 @@ def save_sample_data(self):
 
             # ContainerType - not part of sample or AR schema
             if 'ContainerType' in row:
-                obj = self.lookup(('ContainerType',),
-                                  Title=row['ContainerType'])
-                if obj:
-                    gridrow['ContainerType'] = obj[0].UID
+                title = row['ContainerType']
+                if title:
```
xispa: I think would be better to use 
```python
if row.get('ContainerType',None):`
```
instead

PR #1661, opened 2015-10-24
LIMS-2081: WorkflowException after editing invalid ARImport.

rockfruit:
Silently ignore workflow exceptions; if the validate transition is not
available, there should be some message explaining why to the user.


xispa:
Why not to, at least, log the excpetion message to the debug log?

rockfruit:
The try/except was lazy, now we're just respecting the transition guard.

PR #1662, opened 2015-10-24
LIMS-1993: Secondary AR form *_uid handling for references

rockfruit:
The last fix.

 1 tests with 0 failures and 0 errors in 47.314 seconds.


rockfruit:
@xispa thanks, title and issue are correct this time.

PR #1668, opened 2015-11-02
LIMS-2123: Fix ajax_categories in ServicesWidgets

rockfruit:


rockfruit:
Ran 1 tests with 0 failures and 0 errors in 25.933 seconds.

Espurna:
I don't know if it fails with every listing, but if you try to sort  Worksheet Template's analysis the system returns the view again.

captura de 2015-11-03 18 10 15

PR #1670, opened 2015-11-07
Return specs of each analysis to report

henriquechehad:
Added the specs_dict to return specs informations of each analysis to report

Usage:
Added the var specs_dict returning specs informations of each analysis to report

Usage:
\<span tal:condition="analysis/specs_dict/rangecomment | nothing">
\<span tal:content="structure analysis/specs_dict/rangecomment | nothing">\</span>
\</span>


xispa:
Thanks @henriquechehad , i guess you can obtain the analysis specs by using 'specs' dict key already, see line 551. Why it doesn't work for you? Am I missing something?

henriquechehad:
The specs var is going empty (value: {}) for this report 

I didn't located the var "rangecomment" in another var too.

Example:
https://gist.github.com/henriquechehad/c92962ffaa6dc604d96c

PR #1674, opened 2015-11-10
WINE-71 Instrument. BBK WS export to FIA fails

Espurna:
[JIRA](https://jira.bikalabs.com/browse/WINE-71)

I don't know why Xispa set "DataInterfaceOptions" and" DataInterface" as not visible [here](https://github.com/bikalabs/Bika-LIMS/commit/58ef0fa48f0c1755d7640961513754d20319fa91#diff-2fb6d23392fdd7c495dcccf7aec97f3fR100).

To use the Export system the user has to select the Export instrument interface he/she wants to use from the instrument edition page.

### DIFF
--- a/bika/lims/content/instrument.py
+++ b/bika/lims/content/instrument.py
@@ -115,18 +115,18 @@ schema = BikaFolderSchema.copy() + BikaSchema.copy() + Schema((
         ),
     ),
 
-    #TODO: To be removed?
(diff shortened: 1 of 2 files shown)


xispa:
You're right @Espurna . I set those options to visible=false when I was developing IMM's module and at that time there was no option to export in worksheets I guess.

PR #1677, opened 2015-11-11
WINE-76 WineScan FT120 Import warnings incorrect?

Espurna:
[JIRA](https://jira.bikalabs.com/browse/WINE-76)

I wanted to improve the interface because it is complicated to understand, but I cannot find out how the .csv works, why there is a calibration line every four lines? Which columns should be saved? 

I don't know/understand the utility of _extractrowbycalibration function. 

I know that I did it, but I can't remember how the .csv should be interpreted (Mental note: write clear comments!!!!)

### DIFF
--- a/bika/lims/exportimport/instruments/foss/winescan/ft120.py
+++ b/bika/lims/exportimport/instruments/foss/winescan/ft120.py
@@ -133,15 +133,18 @@ class WinescanFT120CSVParser(WinescanCSVParser):
                 # 2/11/2005 13:33 PM
                 from datetime import datetime
                 dtobj = datetime.strptime(dtstr, '%d/%m/%Y %H:%M %p')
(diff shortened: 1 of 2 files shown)


xispa:
Ummm.... there are two sample files for this instrumet in the source code:
https://github.com/bikalabs/bika.lims/blob/master/bika/lims/exportimport/instruments/foss/winescan/samples/ft120.csv
https://github.com/bikalabs/bika.lims/blob/master/bika/lims/exportimport/instruments/foss/winescan/samples/ft120_full_qcsamples_small.csv

The latter contains this "Calibration" line you mention. Seems to me that results below this "Calibration" header are meant to be calibration tests and this header is used by the parser to generate an index to "identify" to which calibration set each analysis is from (key `Calibration` and value `self._calibration`). Since the importer already takes into account internal calibration tests and results for QC analyses, I think that all this "Calibration" stuff could be omitted.

PR #1680, opened 2015-11-13
LIMS-2133 Once in a while, specs var is going empty in results reports

xispa:
Relates to https://github.com/bikalabs/Bika-LIMS/pull/1670
The AR-specific specs were not taken into account in the reports.
@henriquechehad , with this PR, the specifications are set correctly to `specs` dict key, so `specs_dict` key has been removed.


rockfruit:
:boom:

henriquechehad:
:+1:

PR #1690, opened 2015-11-22
LIMS-2062: Hide cancelled ARs in listing views

rockfruit:


bika/lims/browser/analysisrequest/analysisrequests.py
```diff
@@ -136,9 +137,8 @@ def __init__(self, context, request):
         self.review_states = [
             {'id': 'default',
              'title': _('Active'),
-             'contentFilter': {'cancellation_state': 'active',
-                              'sort_on': 'created',
-                              'sort_order': 'reverse'},
+             'contentFilter': {'sort_on': 'created',
+                               'sort_order': 'reverse'},
```
xispa: Not in master yet, but keep in mind that bika_listing (in development) now automatically sorts by created and in reverse order by default if these params are not set in contentFilter, see:
https://github.com/bikalabs/bika.lims/pull/1686/files#diff-05737f388606e64fa9543e7eea467c36R520
So, we'll not need this anymore unless you want to sort in some other way by default.

PR #1695, opened 2015-11-22
LIMS-2156: Ignore blank index values when calculating ReferenceAnalysesGroupID

rockfruit:


xispa:
For this one, you deserve a beer!

PR #1697, opened 2015-11-22
Expose filter_combogrid from AnalysisRequestAddByCol publicly

xispa:
Required by https://github.com/bikalabs/bika.health/pull/180

Exposed the `filter_combogrid` publicly, accessible via:

```javascript
   window.bika.lims.AnalysisRequestAddByCol.filter_combogrid(element, filterkey, filtervalue, querytype);
```

Although we'd expect that adding the `has_combogrid_widget` wouldn't have any effect, sometimes the referencewidgets don't work properly because of this: while the grid is displayed succesfully, the selected item is not grabbed into the field itself.


rockfruit:
Thank you kindly

PR #1704, opened 2015-11-25
LIMS-1774 Shiny graphs for result ranges

xispa:
Added js for generating result ranges graphs, as well as a new report example.

Adding something like the following in any view will generate the graph automatically:
```html
<div class="range-chart" data-specs="[15,10,20,10]"></div>
```

where the attribute ``data-specs`` is [result, min, max, %error]

The graph gets scaled automatically with the size of the div, so
```html
<div class="range-chart" data-specs="[15,10,20,10]" style="width:200px"></div>
```
will generate a 200px-width graph

ranges2


xispa:
report_sample

PR #1706, opened 2015-11-25
LIMS-2067 Prevent initial method/instrument query for each analysis

rockfruit:
When I load worksheets after applying this PR to master, I notice that the Method and Instrument fields are already completed in the same way as they are when this code is invoked on first render.

@xispa please tell me if I'm not understanding something, or, if there are some cases where the fields must be set differently, so that I can understand what changes need to be made!


xispa:
Umm... don't think this PR fixes [LIMS-2067](https://jira.bikalabs.com/browse/LIMS-2067). Maybe related?

PR #1708, opened 2015-11-26
LIMS-2143 Statements vs Invoices

Espurna:
[JIRA](https://jira.bikalabs.com/browse/LIMS-2143)

@rockfruit I don't know how to change the name in the navigation bar in a installed instance...
It works in a new bika site.

### DIFF
--- a/bika/lims/profiles/default/structure/invoices/.properties
+++ b/bika/lims/profiles/default/structure/invoices/.properties
@@ -1,4 +1,3 @@
 [DEFAULT]
-description = 
-title = Invoices
-
+description =
+title = Statements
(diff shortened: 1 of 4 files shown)


bika/lims/profiles/default/structure/invoices/.properties
```diff
@@ -1,4 +1,3 @@
 [DEFAULT]
-description = 
-title = Invoices
-
+description =
+title = Statements
```
rockfruit: This creates an object who's title attribute is set to 'Statements'.  Beyond that point, the folder is just a piece of Plone content, and so exempt from the normal Plone translation machinery.  The title of the object must be changed using the Plone or Zope UI, or from python code with something like `portal.invoices.title = 'Something'` in an upgrade-step.

Espurna:
ok @rockfruit , I'll try to change the title again.Thank you for the help.

PR #1712, opened 2015-11-30
WINE-81: use standard jquery to parse rows for show_more button

rockfruit:
XML parsing fails when XML is not valid, and it seems that the HTML returned by bika_listing is not always valid XML.  parseHTML is not implemented in the current jquery.

A simple $() works.


xispa:
Thanks!

PR #1718, opened 2015-12-02
HEALTH-331: Sample Preparation Workflow

rockfruit:
Adds an ISamplePrepWorkflow interface which allows objects to have an additional user-created workflow inserted into the workflow chain at runtime.


rockfruit:
https://jira.bikalabs.com/browse/LIMS-1641 - Data capture during sample_prep state (results for some analyses?  Field values?)

https://jira.bikalabs.com/browse/HEALTH-331 - Some advanced features

PR #1737, opened 2016-01-08
WINE-111: Do not attempt detailed formatting of 'nan' results

rockfruit:
in results formatting and decimal point routines, use math.isnan to prevent errors


rockfruit:
tests are missing because I have no idea what causes this.  I've tested it manually in error situations and it works a charm

xispa:
Not sure at all, but maybe "NaN" results are caused because of [LIMS-2175](https://jira.bikalabs.com/browse/LIMS-2175)

PR #1743, opened 2016-01-12
WINE-114: Failed to view published AR (cannot retrieve analysis attachment icon)

rockfruit:


rockfruit:
I missed some; it seems that for certain users, getBestIcon, which calls getContentType, always fails.  Client contacts cannot view anything with getBestIcon inside.

PR #1745, opened 2016-01-13
LIMS-122: Transposed worksheet analysis view fails on blank rows

rockfruit:


xispa:
Thanks!

rockfruit:
It's the first and only bug in that code :)

PR #1751, opened 2016-01-19
LIMS-2194 Error when submitting a result

Espurna:
[JIRA](https://jira.bikalabs.com/browse/LIMS-2194)
I couldn't reproduce the issue, but I wrote some preventive code. Also, I took the freedom to prevent the analysis preselection when the result box is empty.


rockfruit:
Thanks,
I also found this error last week, when the AR specifications are disabled.  We should never assume that a spec is present for any analysis...

PR #1753, opened 2016-01-19
LIMS-1521: Date verified column for AR lists

pakhandibaba:


bika/lims/browser/analysisrequest/analysisrequests.py
```diff
@@ -106,6 +107,9 @@ def __init__(self, context, request):
                                'toggle': SamplingWorkflowEnabled,
                                'input_class': 'datepicker_nofuture',
                                'input_width': '10'},
+            'getDateVerified': {'title': _('Date Verified'),
+                                'index': 'getDateVerified',
```
rockfruit: there is no index getDateVerified - you should remove this line for correctness.

PR #1761, opened 2016-02-03
Lims-2091 More Colums added in Invoice Batch

pakhandibaba:


bika/lims/browser/batchfolder.py
```diff
@@ -112,6 +117,9 @@ def folderitems(self):
             items[x]['Title'] = title
             items[x]['replace']['Title'] = "<a href='%s/%s'>%s</a>" % (items[x]['url'], 'analysisrequests', title)
 
+            items[x]['Client'] = obj.getClient().Title()
```
rockfruit: This will break if there is no client.  Do it in two steps:
    client = obj.getClient()
    if client:
        items[x]['Client'] = client.Title()

bika/lims/browser/invoicebatch.py
```diff
@@ -70,7 +88,17 @@ def folderitems(self, full_objects=False):
             )
             item['replace']['id'] = number_link
             item['client'] = obj.getClient().Title()
```
rockfruit: This also needs to be done with a check to see that the client actually exists, before accessing it's Title and absolute_url() methods.

bika/lims/browser/invoicebatch.py
```diff
@@ -70,7 +88,17 @@ def folderitems(self, full_objects=False):
...
+            item['replace']['client'] = "<a href='%s'>%s</a>" % (
+                obj.getClient().absolute_url(), obj.getClient().Title()
+            )
+            item['email'] = obj.getClient().getEmailAddress()
+            item['replace']['email'] = "<a href='%s'>%s</a>" % (
+                'mailto:%s' % obj.getClient().getEmailAddress(), obj.getClient().getEmailAddress()
```
rockfruit: Here, the same consideration.

bika/lims/browser/invoicebatch.py
```diff
@@ -69,8 +87,24 @@ def folderitems(self, full_objects=False):
...
+                    obj.getClient().absolute_url(), obj.getClient().Title()
+                )
+
+                item['email'] = obj.getClient().getEmailAddress()
+                item['replace']['email'] = "<a href='%s'>%s</a>" % (
+                    'mailto:%s' % obj.getClient().getEmailAddress(), obj.getClient().getEmailAddress()
+                )
+                item['phone'] = obj.getClient().getPhone()
+            else:
+                item['client'] = ''
```
rockfruit: I don't have collaborator access to your fork of bika.lims, so I can't modify this PR directly.  Each column must have a representation in the item{} dictionary.  This else clause should contain:

```
            else:
                item['client'] = ''
                item['email'] = ''
                item['phone'] = ''
```

xispa:
This PR is not a fix for BL3.1 series, so it should go to bikalabs:master instead. Closing.

PR #1763, opened 2016-02-26
LIMS-2208 WinescanCSVParser class instance variable misspelling

trendsetter37:
Fixed bug in `bika.lims.exportimport.instruments.foss.winescan.WinescanCSVParser`

The `self.currentheader` instance variable is declared with all lowercase inside of the constructor.

 ``` python
def __init__(self, csv):
    InstrumentCSVResultsFileParser.__init__(self, csv)
    self.currentheader = None
```

This is consistent throughout the class but with two exceptions.

Lines 26-29

``` python
if not resid:
    self.err("No Sample ID found", numline=self.num_line)
    self.currentHeader = None
    return 0
```

and 

Lines 83-86
 ``` python
self._addRawResult(resid, outvals, True)
self.currentHeader = None
return 0
```
I could not find a reference to `self.currentHeader` (camelcase) in any of the parent classes so I assume this was not intended.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,7 +1,7 @@
 Bika LIMS
 =========
 
-v3.1.12 (unreleased)
+v3.1.11 (unreleased)
 
 The meaning of Gaob
 -------------------
--- a/README.rst
+++ b/README.rst
@@ -1,7 +1,7 @@
 Bika LIMS
 =========
 
-v3.1.11 (unreleased)
+v3.1.12 (unreleased)
 
 The meaning of Gaob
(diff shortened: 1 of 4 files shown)


Espurna:
Please read the comment in your [other pull request](https://github.com/bikalabs/bika.lims/pull/1764)

Why are you changing the version? The next version will be 3.2 I guess, all 3.1.X new fixes should go to support/3.1 without changing the version name.

xispa:
Thanks!

PR #1764, opened 2016-03-01
LIMS-2207 bugfix indentation

trendsetter37:
Fixes failed client starts caused by indentation in `bika.lims.browser.analysisrequest.analysisrequests.py` on Line 660 when using bika.lims source in develop mode.

### DIFF
--- a/bika/lims/browser/analysisrequest/analysisrequests.py
+++ b/bika/lims/browser/analysisrequest/analysisrequests.py
@@ -657,7 +657,7 @@ class AnalysisRequestsView(BikaListingView):
         item['SamplingDate'] = self.ulocalized_time(samplingdate, long_format=1)
         item['getDateReceived'] = self.ulocalized_time(obj.getDateReceived())
         item['getDatePublished'] = self.ulocalized_time(obj.getDatePublished())
-            items[x]['getDateVerified'] = getTransitionDate(obj, 'verify')
+        items[x]['getDateVerified'] = getTransitionDate(obj, 'verify')
 
         deviation = sample.getSamplingDeviation()
         item['SamplingDeviation'] = deviation and deviation.Title() or ''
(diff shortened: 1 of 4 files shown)


Espurna:
A good PR with a good commend, but we usually relate the PR with its issue in [JIRA](https://jira.bikalabs.com/secure/Dashboard.jspa).
Please read [this](https://github.com/bikalabs/bika.lims/wiki/Bika-LIMS-Developer-Guidelines#bika-lims-approach-to-branches-and-pull-requests) documentation about how we work with pull requests :)

I will not close the issue, but please, would you mind to change the title and insert the jira issue? 

It would be wonderful if all of us try to keep an “standard” to create pull requests. Maybe at first it looks like something pity, but I can assure you to keep a well structured pull request agreement is very helpful when it becomes to track the causes of a new issue.

trendsetter37:
@Espurna I don't think I have permission to add labels to the PR

Espurna:
Ok, don't worry about labels, you've done a nice PR! A little fix but documented and reported excellently, we need more of these

bika/lims/browser/analysisrequest/analysisrequests.py
```diff
@@ -657,7 +657,7 @@ def folderitem(self, obj, item, index):
         item['SamplingDate'] = self.ulocalized_time(samplingdate, long_format=1)
         item['getDateReceived'] = self.ulocalized_time(obj.getDateReceived())
         item['getDatePublished'] = self.ulocalized_time(obj.getDatePublished())
-            items[x]['getDateVerified'] = getTransitionDate(obj, 'verify')
+        items[x]['getDateVerified'] = getTransitionDate(obj, 'verify')
```
xispa: Actually, this line should be:
```
item['getDateVerified'] = getTransitionDate(obj, 'verify')
```
Please, @trendsetter37 do this change and submit the PR again!.
We did some refactoring regarding to bikalisting machinery recently: #1654 

Right now, ``folderitem`` service (without the character 's' at the end) is fired directly by BikaListing when rendering lists of items, so we no longer need to iterate over all items again as we did in the past. Nevertheless, most of the BikaListing child classes still need to be refactored to fit with this new behavior.

trendsetter37:
@xispa Just to clarify you would like a new, separate PR, rather than the addition commit? And thanks @xispa & @Espurna for the help I am still new to this but it's fun!

PR #1777, opened 2016-03-11
Travis ci

rockfruit:


xispa:
uhu... automated tests with Travis CI.... this sounds very goood!!!!

bika/lims/tests/test_limitdetections.py
```diff
@@ -70,7 +70,7 @@ def test_ar_manage_results_detectionlimit_selector_manual(self):
              'manual'            : False,
              'input'             : '15',
              'expresult'         : 15.0,
-             'expformattedresult': '15',
+             'expformattedresult': '15.00',
```
xispa: By the way... I suppose all these failed tests with "15.00" are because the value for the field "Precision" is 2 for all the analyses involved. Although I think this is something already covered in another unit test which is focused to test precision changes (fixed and calculated from uncertainties) precisely, remind me to add this extra-condition in these tests too.

PR #1791, opened 2016-03-24
LIMS-2223 Saving a recordswidget as hidden fails

Espurna:
[JIRA](https://jira.bikalabs.com/browse/HEALTH-521)

When a recordswidget's visibility is defined as hidden '...visible={'view': 'hidden', 'edit': 'hidden'},...' the template displays it as an input field with the attribute 'value' as a string 'value="[{:},{:}]"'. This makes the system save the content of the widget as the string instead of a dictionary inside a list, so we need to check if the variable contains a python object as a string.


rockfruit:
What view were you using to see this weird behaviour?  I would like to find out more about it.

Espurna:
https://github.com/nmrl/bika.nmrl/commit/2bc7a47610003017dadc7b65051ed96cb0f34dc3#diff-befb93be018b4ed8b99b33af900388ddR41 This is what I was doing.

PR #1802, opened 2016-05-06
HEALTH-569 Bar code printing not working on sample registration

xispa:
When bar codes are set to be printed at Sample Registration the bar code is not been printed.

Regression due to [LIMS-1539 Printable Worksheets. In both AR by row or column orientations](https://github.com/bikalabs/bika.lims/pull/1434), missed to apply changes in ar-add.js


rockfruit:
this goes back for ages!

xispa:
hehe... since Feb'15

PR #1805, opened 2016-05-11
HEALTH-394 Sample sticker layout. 2 new sticker layouts, 2 stickers per row

xispa:
Added two sticker layouts:
- "Code 93 2x38mm": two 38x18mm stickers per row. Sample ID, Client Sample ID
- "Code 93 2dx38mm": same as above, but with duplicated sticker per row

barcodes4

barcodes5


rockfruit:
merged: failing tests are unrelated, and fixed in another PR.

PR #1823, opened 2016-08-15
Feature/LIMS-2270 Metler Toledo DL55

ramonski:
## Instrument Interface: Metler Toledo DL55

https://jira.bikalabs.com/browse/LIMS-2270

Initial implementation of the instrument including a robot test.
Mappings and value handling needs to be verified, so no merging yet.


lemoene:
**Note misleading issue title**
This is not the LabX' file format, that costs an additional $6000 as per standard Instrument middle ware rip-off, but a file made up by the analyst...

Will be updating the issue with a truer file now

rockfruit:
I've assigned this PR and the issue to myself.

PR #1829, opened 2016-09-01
Setup-data importer: allow analysisservice title or keyword, interchangeably

rockfruit:
This is quite a common issue when creating setupdata sheets - now, get_object allows the AnalysisService title to be a keyword or a title.

### DIFF
--- a/bika/lims/exportimport/setupdata/__init__.py
+++ b/bika/lims/exportimport/setupdata/__init__.py
@@ -267,6 +267,10 @@ class WorksheetImporter:
             logger.info("More than one object found for %s" % contentFilter)
             return None
         elif len(brains) == 0:
+            if portal_type == 'AnalysisService':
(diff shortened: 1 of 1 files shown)


xispa:
@rockfruit code seems ok to me, but..... tests failing? (and not the 'action_id' kind-of-error).

rockfruit:
@xispa the failures we're seeing in travis are unrelated to this - and I have a local branch in which these failures (and other testing issues) are fixed! (I'll submit a good fixed-tests PR soon).  They are related to the history-aware reference field's getter being fired when objects are still in an early state, still being created.  This means that the accessor can't use 'instance', and for the case of tests I have changed this to use zope.component.hooks.getSite() temporarily.  I will probably replace this with a simple boolean check though, since I don't like using getSite in a getter that's used so often.

PR #1831, opened 2016-09-04
LIMS-2325 Allow SampleTypes to be linked with Client Sample Points

rockfruit:


xispa:
In ITS: "LIMS-2325 Editing and saving a Sample Type causes it to lose its Sample Points"

PR #1832, opened 2016-09-05
LIMS-2323 Stickers print using PDF

rockfruit:
(cherry picked from commit 68512870e17de90f40b13359e512553c65dfad2f)
(cherry picked from commit b272d0e14215f0ff0d6b0657df1432ed9c9179d6)

if autoprint=1, then the PDF will be generated immediately - the preview window allows JS to create barcodes, but does not give user time to select another template.  I think that's okay for most uses.

### DIFF
--- a/bika/lims/browser/analysisrequest/workflow.py
+++ b/bika/lims/browser/analysisrequest/workflow.py
@@ -248,7 +248,7 @@ class AnalysisRequestWorkflowAction(WorkflowAction):
         if trans and 'receive' in self.context.bika_setup.getAutoPrintStickers():
             transitioned = [item.id for item in items]
             tmpl = self.context.bika_setup.getAutoStickerTemplate()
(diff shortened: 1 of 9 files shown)


rockfruit:
@xispa I'm not sure about the autoprint feature here.   the autoprint=1 key causes the PDF to be rendered immediately and sent to the browser, but it is then not actually printed.  Additionally, the user must then use the back button to return, and this does not re-load the page they came from.

SO, this probably needs some more work before merging - I'm flagging as not ready.

xispa:
Thanks. Need to check when the autoprint feature is used and review the code. Maybe showing a browser window dettached (popup) containing the generated pdf only would be a better approach, so the user won't need to press the "go back" button. Assigned to me.

ramonski:

ramonski:

PR #1837, opened 2016-09-12
An error in initializeInstrumentsAndMethods:

rockfruit:
The selector '#lab_analyses #item_data, #analyses_form #item_data' does not appear in AR manage_results tabs, and it's absence causes an error that stops JS processing in AR view/edit/manage_results forms.

@xispa I think the 'fix' I applied here may cause issues where this code is expected to work in AR context?

Please verify it!  Thanks.

### DIFF
--- a/bika/lims/browser/js/bika.lims.worksheet.js
+++ b/bika/lims/browser/js/bika.lims.worksheet.js
@@ -367,7 +367,10 @@ function WorksheetManageResultsView() {
         // we'll need them to retrieve all the IMM constraints/rules to be
         // applied later.
         var dictuids = $.parseJSON($('#lab_analyses #item_data, #analyses_form #item_data').val());
-        $.each(dictuids, function(key, value) { auids.push(key); });
+        // in AR manage_results, the above selector doesn't find #item_data.
+        if (dictuids){
+            $.each(dictuids, function(key, value) { auids.push(key); });
+        }
 
         // Retrieve all the rules/constraints to be applied for each analysis
         // by using an ajax call. The json dictionary returned is assigned to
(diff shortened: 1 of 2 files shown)


xispa:
Mmm... only used to retrieve the analysis UIDs listed in the analyses table. Thought this hidden field was always filled in the template, cause is always declared in the pt:

- analysisrequest_analyses: https://github.com/bikalabs/bika.lims/blob/5e78bc9df2df9d2197221b8755ecf328e269d17d/bika/lims/browser/analysisrequest/templates/analysisrequest_analyses.pt#L92
- bika_listing_table (used by the worksheet's default view I guess): https://github.com/bikalabs/bika.lims/blob/5e78bc9df2df9d2197221b8755ecf328e269d17d/bika/lims/browser/templates/bika_listing_table.pt
- analyses_transposed: https://github.com/bikalabs/bika.lims/blob/5e78bc9df2df9d2197221b8755ecf328e269d17d/bika/lims/browser/worksheet/templates/analyses_transposed.pt

Thanks, I'll figure out what is going on.

rockfruit:
#item_data is defined in these cases, but the selector fails because the outer div is not one of #lab_analyses or #analyses_form.

Would be be acceptable to reduce the selector to simply say $('#item_data')?

xispa:
As far as "item_data" is a so generic name that would be used in different scenarios, I'd thought it was better to be more explicit.

In any case, yes, it will work (I've done a quick search form "item_data" in the repos and seems that is only used for analyses tables). Please confirm!

rockfruit:
using just '#item_data' appears to fix the problem nicely,

I'm almost certain that #item_data is used only by bika (there were no instances of the string 'item_data' found in my local Plone instance)

xispa:
Thanks!
btw, "item_data" is meant to be a more easier way to get analyses info from js, to prevent the bizarre way of getting information browsing through the tr/tds and allowing us to minimize the size (kb.) of html by removing unnecessary attributes from the pt.

rockfruit:
OK that's that then, please accept at your leisure.

Thanks!

PR #1844, opened 2016-09-14
LIMS-2391: Use source analysis's sample ID in duplicate analysis IDs

rockfruit:
@xispa any reason we used the samplepartition ID originally?


xispa:
Have no idea... I think using SampleID instead is better. Maybe I'm wrong, but I have the premonition that this change will require mode modifs in the code than only those in the commit d8283c8 . Did all tests passed in your dev environment?

rockfruit:
passed tests and working in the lab.

rockfruit:
I also did a search for 'D' in all code and I can't find any place where there be a problem.

xispa:
> passed tests and working in the lab.

Awesome. Nothing more to say rather accept the PR ;)

bika/lims/content/worksheet.py
```diff
@@ -329,17 +328,16 @@ def addDuplicateAnalyses(self, src_slot, dest_slot):
...
             # the same Reference Sample and same Worksheet)
-            # https://github.com/bikalabs/Bika-LIMS/issues/931
             if not refgid and not analysis.portal_type == 'ReferenceAnalysis':
-                part = analysis.getSamplePartition().id
+                prefix = analysis.aq_parent.getSample().id
                 dups = [an.getReferenceAnalysesGroupID()
                         for an in self.getAnalyses()
                         if an.portal_type == 'DuplicateAnalysis'
-                            and an.getSamplePartition().id == part]
+                            and an.getSamplePartition().id == prefix]
```
xispa: umm... wait, don't understand how this line can work unless ```getSamplePartition()``` returns the Sample, which in any case would be wrong.

xispa: In fact, if you try to add more than one duplicate for the same analysis, I expect that both duplicates will have the sample ReferenceAnalysesGroupID, which shouldn't be

rockfruit: Hum, you are right.  How did I miss it?

PR #1850, opened 2016-09-21
LIMS-2346 Add the time to sampled dates

rockfruit:
SamplingDate and DateSampled now use a jquery datetimepicker.

### DIFF
--- a/bika/lims/browser/sample/view.py
+++ b/bika/lims/browser/sample/view.py
@@ -106,13 +106,13 @@ class SamplesView(BikaListingView):
                       'toggle': False},
             'SamplingDate': {'title': _('Sampling Date'),
                                 'index': 'getSamplingDate',
-                                'input_class': 'datepicker autosave',
+                                'input_class': 'datetimepicker autosave',
                                 'input_width': '10',
                                 'toggle': True},
             'DateSampled': {'title': _('Date Sampled'),
                                'index':'getDateSampled',
                                'toggle': SamplingWorkflowEnabled,
-                               'input_class': 'datepicker_nofuture autosave',
+                               'input_class': 'datetimepicker_nofuture autosave',
                                'input_width': '10'},
             'getSampler': {'title': _('Sampler'),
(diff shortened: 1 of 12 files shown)


bika/lims/browser/sample/view.py
```diff
@@ -106,13 +106,13 @@ def __init__(self, context, request):
...
                                 'index': 'getSamplingDate',
-                                'input_class': 'datepicker autosave',
+                                'input_class': 'datetimepicker_nofuture autosave',
                                 'input_width': '10',
                                 'toggle': True},
             'DateSampled': {'title': _('Date Sampled'),
                                'index':'getDateSampled',
                                'toggle': SamplingWorkflowEnabled,
-                               'input_class': 'datepicker_nofuture autosave',
+                               'input_class': 'datetimepicker_nofuture autosave',
```
xispa: I see "Date Sampled" didn't allow past/present dates before, so is more a question rather than an error to be addressed before accepting the pull request:

Is strictly necessary to not allow a date from the past in "DateSampled" field?. Also, I'd say that the analyst might be able to enter a previous date than SamplingDate.... specially if we include time. Could happen the sampler rearranged his/her agenda on-trip and decided to go for the sample 2h before. Could happen the client creates the Analysis Request once the Sample has already been sampled. 
@lemoene ?

Also note that the field DateSampled in AR hasn't this nofuture restriction: https://github.com/bikalabs/bika.lims/pull/1850/files#diff-c66d564cb5daa751e20534777e638781L44

rockfruit: It's allowing past/present dates (anything up to the current minute), but it is not allowing future time to be used as "DateSampled".  This prevents a condition where the sample is flagged as "DateSampled" of tomorrow, meanwhile it has been received in the lab today.

I cannot imagine why anyone would want to use the "DateSampled" field for a future date - that is why we have the SamplingDate field.

xispa: I see, my interpretation was wrong... Working at night is not healthy. Thanks!

PR #1851, opened 2016-09-21
LIMS-2404: Batch AR list only allows sampling if Sampler and Date Sampled are entered

rockfruit:


bika/lims/browser/bika_listing.py
```diff
@@ -166,6 +166,9 @@ def __call__(self):
         action, came_from = self._get_form_workflow_action()
 
         if action:
+            # bika_listing sometimes gives us a list of items?
+            if type(action) == list:
+                action = action[0]
```
xispa: I suggest to move https://github.com/bikalabs/bika.lims/blob/master/bika/lims/browser/bika_listing.py#L84
to line 95 (from bika_listing) instead. Otherwise, we are forced to do this check everywhere.

rockfruit: Thanks.

PR #1860, opened 2016-10-02
ScheduleSamplingEnabled and DefaultCountry are not imported from setupdata

xispa:
Added 'ScheduleSampling' and 'DefaultCountry' in setupdata import and test.xlsx


xispa:
BTW, to keep track of the changes done in the xlsx, maybe we could save a JSON file that represents the xlsx in the repos and later, let plone/bika.lims convert back that JSON to xlsx.
Set the option "Output Format" to JSON and drop the test.xlsx file here: http://oss.sheetjs.com/js-xlsx/ ....  you get a beautiful JSON ready to make git happy. The same tool can be used to generate the xlsx back, but probably there are other tools written in python out there that can do the same too.

PR #1870, opened 2016-11-23
Selecting one analysis for verification from AR results view, verifies all

xispa:


bika/lims/browser/analysisrequest/workflow.py
```diff
@@ -463,7 +463,10 @@ def workflow_action_publish(self):
     def workflow_action_verify(self):
         # default bika_listing.py/WorkflowAction, but then go to view screen.
         self.destination_url = self.context.absolute_url()
-        return self.workflow_action_default(action='verify', came_from='edit')
+        action, came_from = WorkflowAction._get_form_workflow_action(self)
+        if type(came_from) in (list, tuple):
```
rockfruit: This happens when multiple parameters have the same name in the request; plone creates a list instead of allowing them to override each other.  I just haven't been able to see how the extra parameter is inserted...

PR #1873, opened 2016-12-01
LIMS-2455: (Lab)Contact/User Linkage Behavior

ramonski:
## Description

System Users can be assigned to a `Contact` or `LabContact`. The `login_form` allows now to *link* or *unlink* an existing User via a small search form or create a new user, which will be automatically linked to the (Lab)Contact.
If the `Contact` or `LabContact` is deactivated and has a linked User, the User cannot log into the system anymore and unlinking the user from the (Lab)Contact is disabled.

See: https://jira.bikalabs.com/browse/LIMS-2455

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -30,3 +30,4 @@ Jonas Daniel <jonas@ogaard.no>
 Luke Williams <luke@iquidus.co.nz>
 Manav Batra <manavbatra@outlook.com>
 Aman Arora <aman2arora2@gmail.com>
+Ramon Bartl <rb@ridingbytes.com>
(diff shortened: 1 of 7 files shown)


rockfruit:
looking really good, been a long time since someone loved this form.

ramonski:
## Login behavior

The current code denies a login when a contact is deactivated via the `logged_in.cpy` (https://github.com/bikalabs/bika.lims/pull/1873/files#diff-c24c6c27208772e5a53bc607c2ba1721) but it should be improved. E.g. it displays details the username in the upper right corner and instrument calibrations/certificates before the user is logged out.

A better approach would probably involve the coding of a PAS plugin, which uses the `UID` of the linked contact, which is stored as a user property on linking, to gather the active/inactive state of the `LabContact`/`Contact`:
https://github.com/bikalabs/bika.lims/pull/1873/files#diff-8c830fee632bcf8e764447ed52a2f7b1R237

PR #1885, opened 2016-12-08
LIMS-1391 Configurable object identifiers

rockfruit:
Provides an interface 'IHaveIdentifiers' which allows objects providing it to be referred to by multiple configurable IDs.  This can be used to allow objects to have CAS numbers assigned to them, or to allow a Patient record to be looked up via a national ID, driver's license, or hospital record number.

To use this:

- Provide the bika.lims.interfaces.IHaveIdentifiers interface on your object.  These objects will now have a new field added to their schema called 'Identifiers'.  The field contains a list of dictionaries:  {'IdentifierType', 'Identifier', 'Title'}.

- Enter Bika Setup, and select "Identifier Types".  Here you can configure the name and description of all types of identifiers to be used on the system.  You can also specify a list of portal_types which each identifier type can be used to identify.

- A new KeywordIndex index 'Identifiers' is added to portal_catalog, bika_catalog and bika_setup_catalog.

- The field is flagged 'searchable=True' so 'Identifier' values are available in the SearchText ZCFullText index.

- This commit adds an IdentifierType called 'CAS Nr' representing Chemical Abstract Service registry numbers, and applies the IHaveIdentifiers interface to AnalysisServices


xispa:
This one is pretty cool!

PR #1890, opened 2016-12-15
Added i18n attributes to WF Buttons

ramonski:
Changeset for https://jira.bikalabs.com/browse/LIMS-2465


ramonski:
The `plone.pot` file contains msgids for the workflow transition titles, e.g. `republish_transition_title`, which were ostensibly found in `bika.lims/profiles/default/workflows/<workflow_definition_xml_file>`.
However, this directory does not exist anymore in the current codebase, as the worfklow definitions moved to `bika.lims/profiles/default/workflow_csv/<workflow_definition_csv_file>`. 
Thus, the actions of the Bika Listing Table cannot be translated anymore and msgstrs in the po files have no effect:

```
#. Default: "Republish"
#: ../profiles/default/workflows/bika_ar_workflow/definition.xml
msgid "republish_transition_title"
msgstr "Wiederveröffentlichen"
```

I'm not sure, why the translation machinery keeps these orphane `msgstr` entries in the `plone.pot`, but they should be kept in.
If they are removed, the Button titles change to `string:${action/id}_transition_title` and we need another way to extract the transition titles from the `CSV` definitions.

PR #1891, opened 2016-12-15
Multi-verification, but different verifiers

Nihadness:
At the moment, the multi-verification allows the LabManager to set the number of verifications for an analysis are needed before an analysis being transitioned to the 'verified' status. Nevertheless, the same user (with "verifier" privileges) can trigger all the verifications until the analysis get verified.

The system must allow the LabManager to set one of the three options below in analysis service and also in Bika Setup when multi-verification is enabled:
- The analysis can be verified multiple-times by the same verifier
- The analysis can be verified multiple-times by the same verifier, but not consecutively
- The analysis cannot be verified multiple-times by the same user

A selection list in Bika Setup must be added below the option "Allow multi-verification of results", with the options listed above. This selection list will only be displayed if the value for the option "Allow multi-verification of results" is >= 2
A selection list in Analysis Service edit view must be added below the option "Allow multi-verification of results", with the options "System default" (by default, will take the value set in Bika Setup) plus the options listed above. This selection list will be displayed if the value for the option "Allow multi-verification of results" is >= 2. The value set here prevails over the value set in Bika Setup.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -31,3 +31,4 @@ Luke Williams <luke@iquidus.co.nz>
 Manav Batra <manavbatra@outlook.com>
 Aman Arora <aman2arora2@gmail.com>
 Ramon Bartl <rb@ridingbytes.com>
+Nihadness <nmammadli@naralabs.com>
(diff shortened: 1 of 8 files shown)


Nihadness:
I missed something!

PR #1892, opened 2016-12-15
Fixed buildout failure due to CairoSVG 2.0

ryoung29:
Fixed buildout failure caused by the new python 3 requirement in CairoSVG 2.0.

Also, tested installation on Ubuntu 16.04.

### DIFF
--- a/docs/INSTALL.rst
+++ b/docs/INSTALL.rst
@@ -10,7 +10,7 @@ Linux Installation Steps
 
     The following list of packages need to be installed.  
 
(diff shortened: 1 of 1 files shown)


rockfruit:
Hi, thanks,

this is fixed here: https://github.com/bikalabs/bika.lims/blob/master/setup.py#L75

cheers

rockfruit:
Hi, thanks,

this is fixed here: https://github.com/bikalabs/bika.lims/blob/master/setup.py#L75

cheers

PR #1898, opened 2016-12-22
Export lists to CSV or XML

xispa:
Add "Export to CSV" and "Export to XML" buttons in lists to allow the user to export the data to a .CSV or .XML file. The file must display the same items, with the same columns and with the same order as displayed in the list. This functionality doesn't meant to be used for automatic harvesting, rather to allow the user to easy (and fast) export any list to an understandable format, anytime.

export-csv-xml


xispa:
See #1911

PR #1901, opened 2016-12-28
Reflex rules

Espurna:
When the results are introduced, some samples may need to be added to the next available worksheet in order to retest it and confirm the result or maybe the result of a previous analysis should be changed depending on the result of the reflexed analysis. These situations can be caused by the indetermination of the result or by a failed test.

This functionality was designed to bring a high level of flexibility regarding to dynamic and automated creation of analyses, duplicates and other types of transitions without the need of user interaction.

Basic idea:

Each reflex rule have to be bound to an analysis method using the drop-down list. So, the actions defined inside that reflex rule will be triggered only if the analysis is submitted/verified using this method.

Inside the reflex rule the user will be able to add actions for each analysis service belonging to the selected method. In order to improve the flexibility of the system the user can set some conditions over the analysis service. For instance, the user can restrict the trigger of the actions to only those analyses which a reflex rule has been applied to them (taking advantage of the method).

An example of the usage could be: when a result "0" (e.g. Indeterminate) is submitted for an analysis "EID" using a method named "Method-X". Then the reflex-rule criteria is met and two duplicates will be generated and added into the next available worksheet assigned to the analyst specified. Later, if the analyst sets a result for those duplicates and any of the "second-level" conditions are met, the action "set result" will be triggered: a new analysis EID with result "0" (Indeterminate) will be added and set as the definitive result.


bika/lims/content/analysis.py
```diff
@@ -1332,9 +1436,9 @@ def workflow_script_retract(self):
         # Rename the analysis to make way for it's successor.
         # Support multiple retractions by renaming to *-0, *-1, etc
         parent = self.aq_parent
-        analyses = [x for x in parent.objectValues("Analysis")
-                    if x.getId().startswith(self.id)]
         kw = self.getKeyword()
+        analyses = [x for x in parent.objectValues("Analysis")
+                    if x.getId().startswith(kw)]
```
xispa: To be reviewed. IMO, the correct should be:
```
analyses = [x for x in parent.objectValues("Analysis")
                    if x.getId().startswith(self.id)]
```
Note this is triggered during the retraction of an analysis and this is used to generate an autonumeric suffix for the new analysis to be created because of the retraction

bika/lims/content/analysisservice.py
```diff
@@ -1218,6 +1218,22 @@ def getAvailableMethods(self):
 
         return methods
 
+    def getAvailableMethodsUIDs(self):
+        """ Returns the UIDs of the available method.
```
xispa: The content of this service duplicates `getAvailableMethods` https://github.com/bikalabs/bika.lims/blob/master/bika/lims/content/analysisservice.py#L1199.

Much more easier:
```
muids = [m.UID() for m in self.getAvailableMethods()]
```

xispa:
See #1911

PR #1904, opened 2017-01-09
Filter Bar for lists

Espurna:
This is the skeleton of the functionality, a new PR will come with the bar itself (already created and tested in another repo) and the enable/disable button in bika_setup.

This PR should be accepted in order to avoid future merge conflicts, the next PR with the bar will not create any conflict if this PR is accepted before because takes advantage of the functions defined here.

Description from JIRA:

> 
> The user could use the filter to sort either the list of Analysis Requests or the list of “Samples with results”
> Functional description:
> A new functionality in bika listing should be added, this functionality consists on a filtering bar to allow the user filter the list by more than one field or rang of values.
> This filter bar should be located over the current filter buttons that allow the filtering by state in bika listing.
> This new bar should display a “filter” button in order to submit the changes.
> This filter shouldn’t override the filter by state functionality
> Technical approach:
> Add a new tr element inside the “.listing-filter-row table”. The new filter bar will be located here.
> In order to make this bar modifiable in other add-ons, a good approach could be to create a new template only for this td, so it could be modified in other views and addons if needed. So inside BikaListingView, define and create a new template file.
> In Bika LIMS some new class methods may be created in order to make the adaption to other views easier.
> First create a class attribute, this could be a dictionary to store the filter fields, its type and its indexes
> _filter_array = [
> {‘type’:’date_range’, ‘index’: ‘getDateTested’}
> , …]_
> The type defines the widget type for the filter’s field and the index defines a catalog index to query.
> Then a method filter_bar_builder should be created, the template is going to call this method to fill the filter field options.
> Override the method and attribut to define the filters on health
> Filters and views:
> Analysis Request: Filter by date received (range), Date tested (range), Analysis name, of range results, batch.
> Samples: Filter by sample types, Sample condition, Date received (range), batch.

### DIFF
--- a/bika/lims/browser/analysisrequest/workflow.py
+++ b/bika/lims/browser/analysisrequest/workflow.py
@@ -51,7 +51,7 @@ class AnalysisRequestWorkflowAction(WorkflowAction):
             came_from = came_from[0]
         # Call out to the workflow action method
         # Use default bika_listing.py/WorkflowAction for other transitions
-        method_name = 'workflow_action_' + action
+        method_name = 'workflow_action_' + action if action else ''
         method = getattr(self, method_name, False)
         if method:
             method()
--- a/bika/lims/browser/bika_listing.py
+++ b/bika/lims/browser/bika_listing.py
@@ -8,6 +8,7 @@
 import json
 import re
 import urllib
+import copy
 from operator import itemgetter
 
 import App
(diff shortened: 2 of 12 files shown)


xispa:
@Espurna tests failing

bika/lims/browser/analysisrequest/workflow.py
```diff
@@ -51,7 +51,7 @@ def __call__(self):
             came_from = came_from[0]
         # Call out to the workflow action method
         # Use default bika_listing.py/WorkflowAction for other transitions
-        method_name = 'workflow_action_' + action
+        method_name = 'workflow_action_' + action if action else ''
```
rockfruit: 'workflow_action_' + '' will still be an error, should probably abort/exit/return if this happens, with a warning.

bika/lims/browser/bika_listing.py
```diff
@@ -822,12 +847,15 @@ def folderitems(self, full_objects = False):
             show_all = True
         else:
             show_all = False
-
+        contentFilterTemp = copy.deepcopy(self.contentFilter)
```
rockfruit: Thanks for sorting that, I know modifying the default class property was causing issues

PR #1906, opened 2017-01-12
Lims 2477 reference has no dependencies

rockfruit:


xispa:
I see... a Reference Analysis will never have dependencies, a "manual" result is always expected.

PR #1907, opened 2017-01-12
LIMS-2477: getSubmittedBy not available on reference-analysis

rockfruit:
pull #1906 this appears to be related


rockfruit:
Found more of the same type from @Nihadness in december, I'll update this PR in a second

Nihadness:
Before my pull request, it getSubmitedBy() wasn't checked anywhere if it exists or not, so I didn't check as well and just called it... 
Anyways, my question is that, should I always add hasAttirbute() function every time we add new function to the object, because old ones might not have this function? Is there a way to add new functions to old objects as well? Maybe the same thing as reindexing or something like that.

xispa:
@Nihadness . The problem is that in "Analyses" view, any of the objects from types `Analysis`, `ReferenceAnalysis` and `DuplicateAnalysis` can be listed. While `DuplicateAnalysis` inherits from `Analysis` (so acquires the method `getSubmittedBy()`), the objects from type `ReferenceAnalysis` don't inherit from `Analysis`, so they don't have the function `getSubmittedBy()`.

We'll eventually need to make `ReferenceAnalysis` inherit from `Analysis` too and override those functions that don't make sense in that context. Meanwhile @rockfruit and @Nihadness I suggest to add the function `getSubmittedBy()` to `ReferenceAnalysis` too, instead of using `hasattr`

rockfruit:
xispa you're right, the submit transition exists in referenceanalysis_workflow and also in duplicateanalysis_workflow, so these should both be permitted this behaviour.

@Nihadness sorry, didn't mean to lay blame! thanks :P

PR #1908, opened 2017-01-13
LIMS-2480 Fix issues filtering and sorting batch folder listing

rockfruit:
Simply added missing indexes (and sort-index for BatchID)

### DIFF
--- a/bika/lims/setuphandlers.py
+++ b/bika/lims/setuphandlers.py
@@ -616,6 +616,7 @@ class BikaGenerator:
         addIndex(bc, 'getAnalysisService', 'KeywordIndex')
         addIndex(bc, 'getAnalyst', 'FieldIndex')
         addIndex(bc, 'getAnalysts', 'KeywordIndex')
+        addIndex(bc, 'BatchID', 'FieldIndex', 'getBatchID')
         addIndex(bc, 'BatchDate', 'DateIndex')
         addIndex(bc, 'getClientOrderNumber', 'FieldIndex')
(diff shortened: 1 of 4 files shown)


bika/lims/setuphandlers.py
```diff
@@ -616,6 +616,7 @@ def addColumn(cat, col):
         addIndex(bc, 'getAnalysisService', 'KeywordIndex')
         addIndex(bc, 'getAnalyst', 'FieldIndex')
         addIndex(bc, 'getAnalysts', 'KeywordIndex')
+        addIndex(bc, 'BatchID', 'FieldIndex', 'getBatchID')
```
ramonski: Do we need an upgrade step to reindex this index?

ramonski:
Tasks taken from https://jira.bikalabs.com/browse/LIMS-2480:

- [X] Neither the Filters - Open, Closed, Cancelled, nor sorting seems to work.
       -> works

- [X] Title sorts OK on the first click, then doesn't sort on any further clicks.
         -> works

- [ ] Batch ID does not sort at all, same for date and description
       -> does not work on my machine, although the index `BatchID` exists and is populated.

- [X] Default sorting should be reverse Batch ID order? Freshest at the top
         -> works (sorts on created date in reverse order)

- [] The filters on the global list work, but not sorting. Only Date does on the first click.
         -> The WF-Buttons work, but sorting by BatchID is still not functional

I guess we should add an upgrade step to reindex the `BatchID` in the `bika_catalog` tool.

ramonski:

PR #1916, opened 2017-01-25
LIMS-2496 Fix multi-AR publication

rockfruit:
Some performance, formatting, and email fixes for publication of multi-AR and single-AR templates


rockfruit:
wkhtmltopdf requires an X server, xvfb works fine; so the installation docs need to be updated with this info.

wkhtmltopdf doesn't support @page{}, but gives good control over these values when invoking pdfkit.from_*() methods.  Some site-setup values are now required to configure these.

PR #1917, opened 2017-01-25
Solved WF Translation issues and fixed WF Action Buttons in Bika Listings

ramonski:
- Dropped `collective.wtf` CSV WF definitions and all occurances
- Re-added WF definitions in XML to be translatable
- Updated i18n msgids for WF States
- Fixed WF Action buttons in `bika_listing.py` and `bika_listing_table.pt`
- Fixed untranslated WF States in listing tables
- PEP8 and better formatting of  `bika_listing.py` and `bika_listing_table.pt`
- Fixed missing `bikaMessageFactory` import in `workflow.py`
- Updated German translations


ramonski:
Hi Campbell,

I went through all the Workflow XML definitions and formatted them properly and fixed here and there some missing `i18n:attributes="title"`.

I also consolidated all WF State titles, as they were kind of messy. Sadly the Plone WF Action menu still keeps untranslated, so that I had to add all the WF states from `bika.pot` into `plone-manual.pot` to fix this behavior.

The Action Buttons of the listing tables get now as well properly translated and are now functioning again. I also fixed the "WF State" column of the listing table, so that they are now translated as well.

The rest of the changes encompass mainly a dozen `update_translation` runs and some PEP8 and formatting stuff.

It would be awesome if you could review and merge this soon, as the current `master` branch contains the non-functioning workflow action buttons, which I accidently introduced through https://github.com/bikalabs/bika.lims/pull/1890 (sorry about that).

Thanks a lot dude!

PR #1920, opened 2017-02-01
Workflow/Permission Rework

ramonski:
## Description

Major Workflow and Permission overhaul.

The following Views (probably more) are currently accessible by Anonymous:

http://localhost:8080/Plone/bika_setup/bika_labcontacts/base_view
http://localhost:8080/Plone/bika_setup/bika_analysiscategories/base_view
http://localhost:8080/Plone/methods
http://localhost:8080/Plone/bika_setup/bika_departments/base_view
http://localhost:8080/Plone/bika_setup/bika_samplepoints/base_view
http://localhost:8080/Plone/bika_setup/bika_suppliers/base_view
http://localhost:8080/Plone/bika_setup/bika_instruments/base_view
http://localhost:8080/Plone/bika_setup/bika_analysisprofiles/base_view

Some of them are so by purpose, e.g. Anonymous users may see Methods and Services and the Accreditation Logo needs to be accessible from the Portlet on the frontpage for unauthenticated users.

However, others maybe not and this has to be figured out within this PR.

## Things done so far

- Added a Doctest for workflow/permission testing: `bin/test -m test_textual_doctests`
- Removed the secondary `one_state_workflow` for `bika_setup` contents
- Added guard permission `Review portal content` for `bika_inactive_workflow`
- Added explicit permissions (`acquire="False"`) for `View`, `List folder contents` and `Access contents information` in `bika_one_state_workflow`
- Added implicit permissions (`acquire="True"`) for `View`, `List folder contents` and `Access contents information` in `bika_inactive_workflow`


lemoene:
_Anonymous users may see Methods and Services and the Accreditation Logo needs to be accessible from the Portlet on the frontpage for unauthenticated users._

These should be by default visible to anonynmous visitors:

- Methods

- Accreditation page hyperlinked from the logo

- **NB Not services?** afaik. Labmanagers may create Price lists incl AS descriptions to fulfil such a requirement

ramonski:
Please user for testing this command:

```
bin/test -m test_textual_doctests
```

For debugging, it is recommended that you install `pdbpp`, a drop-in replacement for `pdb`
https://pypi.python.org/pypi/pdbpp

This allows you to add a line like this in the `bika.lims/docs/Permission.rst` doctest to get proper debugging possibilities:

```
>>> import pdb;pdb.set_trace()
```

rockfruit:
Some parts of services are made visible to show accredited services
listing. There's probably a better way... Or at least to restrict the
fields that are accessible

ramonski:
Objects of type `AnalysisService` have now `View` and `Access contents information` permissions granted for Anonymous: https://github.com/bikalabs/bika.lims/pull/1920/files#diff-41cc111d5fb99fc3cbcb9e1b1b24490fR443

ramonski:
## Architectural Note

- The `bika_one_state_workflow` sets explicit roles (`acquire="False"`) for the following permissions:
    - Delete objects
    - Modify portal content
    - View
    - Access contents information
    - List folder contents

- The `bika_inactive_workflow` sets implicit roles (`acquire="True"`) for the following permissions:
    - Delete objects
    - Modify portal content
    - View
    - Access contents information
    - List folder contents

- The `BikaSetup` object and its contained objects on the **first level** (`bika_instrument`, `laboratory` etc.) have the `bika_one_state_workflow` assigned

- Objects of the **second level** below `bika_setup` have the `bika_inactive_workflow` assigned

This means, that the custom settings applied in `setuphandlers.py` control the permissions of the lower level objects, e.g. Instruments, Analysis Services etc.
Here it is controlled which objects are accessible by unauthenticated Members.

## Important

Running `Update Security Settings` in `/portal_workflow/updateRoleMappings` will overwrite custom permissions from the `setuphandler`. Reinstallation of `bika.lims` through the quickinstaller is then necessary.

Updating the Permissions will require a **Catalog Rebuild**. Otherwise the listing tables may stay empty.

ramonski:
@xispa, @rockfruit: The current PR works fine with a fresh installation of bika lims, and the described security issues are fixed. I just changed so far the permissions for `bika_setup`, `methods` and `bika_analysisservices` in `setuphandler.py` to be as permissive as possible. The doctest also returns expected results for the objects I test.

ramonski:
## Notes from Slack

ramonski [16:36] 
rockfruit: was there any reason, why some contents had the `bika_one_state_workflow` *and* the `bika_inactive_workflow` set?
I also asked xispa, but we came to the conclusion that it is not needed.
However, I ask you because of the current build failure of the PR: https://travis-ci.org/bikalabs/bika.lims/builds/197672667

rockfruit [16:39] 
because bika_inactive_workflow and bika_cancellation_worflow are 'secondary' workflows, they do not use 'review_state' - but plone requires all objects that subscribe to workflows to have a 'review_state' and fails without it.  SO I copied Plone itself, which also has a one_state_workflow.

[16:40]  
Even if you made cancellation or inactive workflow into the 'primary' (only) workflow for an object, that would be confusing - because then any object that has an extensive primary workflow, and uses cancellation/inactive workflow as secondary, would be using a totally different workflow definition.xml

[16:41]  
(on account of the hard-coded review_state requirement).

[16:41]  
I freakin hate that message btw.  It's a failure of translation in CMFCore, easy to fix, but nobody ever does.  I usually hack it fixed in my local machine, just so that I can debug without having to use PDB

[16:42]  
I don't even know why they bother putting it through i18n - it's a developer level message, it does not need to be german or spanish.

PR #1924, opened 2017-02-02
LIMS-2548: Allow negative numbers in LDL and UDL

rockfruit:


bika/lims/browser/js/bika.lims.site.js
```diff
@@ -340,8 +340,8 @@ function SiteView() {
 
         // Archetypes :float and DecimalWidget inputs get filtered
         $("input[name*='\\:float'], .ArchetypesDecimalWidget input").keyup(function(e) {
```
ramonski: I think we could even drop the whole JS function, as this should be the standard behavior of this field.

PR #1911, opened 2017-01-20
Lots of stuff

xispa:
**NOT READY YET**
Apart from reviewing all works perfectly well, I want to allow the labman to enable/disable the Aggregated list of analyses via Bika Setup, cause this functionality could be annoying if the lab runs a lot of different tests (and with calculations).

**Attention:** Upgrade step 320 is required. This upgrade step can take long, so be patient an plan the upgrade beforehand.

Includes:
- Reflex testing,
- Department filtering,
- Filter-bar on lists,
- Multi-approval of results,
- Rejection reasons,
- Lists export,
- Aggregated analyses view,
- Printing workflow for ARs,
- Analyses figures in dashboard,
- Multi-department assignment to labcontacts,
- Worksheet templates with method enforcement

The following PRs are included in this one:
- #1898 Exported lists to CSV and XML
- #1901 Reflex rules

PR #1869, opened 2016-11-23
Backwards compatibility (multi-approval of results)

xispa:

PR #1928, opened 2017-02-09
Lots of stuff (reflex, department filtering, multi-approval, rejection reasons, etc.)

xispa:
Replaces #1911 

I've reviewed the code thoroughly, I've ran the tests and also tested basic workflows in the browser.
Includes all the work done for NMRL project that can be included in the codebase.

**Attention:** Upgrade step 320 is required. This upgrade step can take long, so be patient an plan the upgrade beforehand.

Includes:
- Reflex testing,
- Department filtering,
- Filter-bar on lists,
- Multi-approval of results,
- Rejection reasons,
- Lists export,
- Aggregated analyses view,
- Printing workflow for ARs,
- Analyses figures in dashboard,
- Multi-department assignment to labcontacts,
- Worksheet templates with method enforcement

The following PRs are included in this one:
- #1898 Exported lists to CSV and XML
- #1901 Reflex rules


bika/lims/browser/widgets/reflexrulewidget.py
```diff
@@ -0,0 +1,569 @@
...
+                'worksheettemplate': worksheettemplate,
+                'analyst': analyst,
+                'setresulton': setresulton,
+                'setresultdiscrete': setresultdiscrete,
+                'setresultvalue': setresultvalue,
+                'an_result_id': local_id,
+                }
+            # Saves the action as a new dict inside the actions list
+            actions_dicts_l.append(action_dict)
+            a_count += 1
```
ramonski: same here

bika/lims/browser/widgets/reflexrulewidget.py
```diff
@@ -0,0 +1,569 @@
...
+        for method in methods:
+            # Get the analysis services related to each method
+            br = method.getBackReferences('AnalysisServiceMethods')
+            analysiservices = {}
+            for analysiservice in br:
+                # Getting the worksheet templates that could be used with the
+                # analysis, those worksheet templates are the ones without
+                # method and the ones with a method shared with the
+                # analysis service.
+                bsc = getToolByName(self, 'bika_setup_catalog')
```
ramonski: better get this tool outside of the loop

bika/lims/browser/widgets/reflexrulewidget.py
```diff
@@ -0,0 +1,569 @@
...
+                'analysisservices': analysiservices,
+                'as_keys': analysiservices.keys(),
+            }
+        # Get the data saved in the object
+        relations['saved_actions'] = {
+            'method_uid': self.aq_parent.getMethod().UID() if
+            self.aq_parent.getMethod() else '',
+            'method_id': self.aq_parent.getMethod().id if
+            self.aq_parent.getMethod() else '',
+            'method_tile': self.aq_parent.getMethod().Title() if
```
ramonski: better get `self.aq_parent.getMethod()` once and store it in a variable

bika/lims/browser/widgets/reflexrulewidget.py
```diff
@@ -0,0 +1,569 @@
...
+            self.aq_parent.getMethod() else '',
+            'rules': self.aq_parent.getReflexRules(),
+            }
+        return json.dumps(relations)
+
```
ramonski: The methods below have all no translated titles in the display lists

bika/lims/browser/widgets/reflexrulewidget.py
```diff
@@ -0,0 +1,569 @@
...
+        :element: a string with the name of the element of the action to
+            obtain: 'analysisservice', 'cond_row_idx', 'range0', 'range1',
+                    'discreteresult', and_or
+        """
+        if isinstance(set_idx, str):
+            set_idx = int(set_idx)
+        if isinstance(row_idx, str):
+            row_idx = int(row_idx)
+        cond = self.getReflexRuleElement(idx=set_idx, element='conditions')
+        return cond[row_idx].get(element, '')
```
ramonski: This fails if `getReflexRuleElement` returns an empty string (see Line 522)

bika/lims/browser/widgets/serviceswidget.py
```diff
@@ -41,9 +44,14 @@ def __init__(self, context, request, field=None, category=None):
         self.show_select_column = True
         self.pagesize = 999999
         self.form_id = 'serviceswidget'
-
         self.show_categories = False
         self.categories = []
+        self.method_filter = None
+        # Filter by method
+        if hasattr(self.context, 'getRestrictToMethod') and\
+                getattr(self.context, 'getMethodUID')():
```
ramonski: Why are we doing here one time `hasattr` and then `getattr`?

bika/lims/content/analysisrequest.py
```diff
@@ -1900,6 +1895,38 @@ def getLate(self):
 
         return False
 
+    def getPrinted(self):
+        """ returns "0", "1" or "2" to indicate Printed state.
+            0 -> Never printed.
+            1 -> Printed after last publish
+            2 -> Printed but republished afterwards.
```
ramonski: Please add variables for these states, that makes the code more readable, e.g.:

`NEVER_PRINTED = "1",  ...`

bika/lims/content/analysisrequest.py
```diff
@@ -1900,6 +1895,38 @@ def getLate(self):
...
+                    return "2"
+        return "0"
+
+    def printLastReport(self):
+        """ Setting Printed Time of the last report, so its Printed value will be 1"""
+        workflow = getToolByName(self, 'portal_workflow')
+        review_state = workflow.getInfoFor(self, 'review_state', '')
+        if review_state not in ['published']:
+            return
+        last_report=sorted(self.objectValues('ARReport'),key=lambda report: report.getDatePublished())[-1]
```
ramonski: do we always have at least one ARReport?

bika/lims/content/bikacache.py
```diff
@@ -0,0 +1,36 @@
...
+
+    # 'Value' is ID of the last created object. Must be inscreased before using.
+    atapi.StringField('Value',default='')
+
+))
+
+schema['title'].widget.visible = False
+
+class BikaCache(BaseContent):
+    """
```
ramonski: I would not recommend this approach as it will clutter our database.
Please consider something like a persistent utility + Annotation Storage to do this job

xispa:
Thanks @ramonski  , don't be sorry to say what you think!. Is always good to receive feedback (either good or bad), is the engine that makes us improve on our skills, or at least, force us to think.

Yes, I agree with you this PR is too much complex, here comes the story: I've always been advocating that, for client-specific projects, the best approach would be to fork from master branch at the beginning and allocate a month in the project timeline for an "alignment" with the head branch every 3 months. This used to work pretty well, but in the "alignment" I planned for the past January, we noticed that more than 200k lines were added and nearly 290k were removed (yea, I know a lot of them were .po files), so I ended up with a lot of merge conflicts when I tried to do atomic PRs against master. I did several attempts before, on December, but after spending a lot of time struggling with conflicts, I finally realized that I had no choice: this amorphous PR was born. It was a decision to forge ahead, with the hope the divergence could be still amended in a near future. My wrong. From Jan until today (1.5 months), more than 85k lines have been added and more than 60k deleted. Bika Health is another story, although no contributions have been done there since Mar 2016 (except a little fix regarding Analysis Specs), I am sure the code there is not compatible with the latest master, neither the health code we've been working on during the last 4 months.

I take note, feel free to close the PR for now. Hopefully, at some point, we'll be able to sanitize all this.

ramonski:
Thanks @xispa, I can completely understand the circumstances that lead to this huge commit.

When I said complex, I meant more the  data structures / lines of code per function / indentation etc. and not the amount of changed lines.

Anyhow, it would be sad if we stop here now, as there are basically a lot of good approaches in this code.

We should rather think how we could split up the functionality into smaller parts and put it into smaller PRs.

Please feel free to contact me for further review or architectural decisions, I'm happy to assist.

xispa:
Appreciated, we'll try to find out a solution.

Btw, the master+develop branches strategy (later hotfix+master and later support+master) we used to have in the past was precisely meant to prevent this type of problems. The idea was to always fork from master on new projects, push fixes and typos only against master, new features and arch restructurations against develop. Merge master into develop regularly and merge master into client-fork regularly too. After ~6 months, sanitize develop, freeze it, rename master to support, rename develop to naster, publish a new release and create a new "develop" branch... And start again. Support branches were meant to have the opportunity to give LTS to old clients or bika custom instances deployed more than 1y before.

Just my two cents!

ramonski:
Thanks for the detailed description – Indeed, that is a good and solid approach, although it introduces some overhead, which needs to be handled by someone.

When I started to contribute to `bika.lims`, I did not meant to drop existing processes, so don't get me wrong on this.

I completely agree with the LTS idea and that the `master` branch needs to be kept "holy", but from my experience it is anyhow better and more maintainable to have a single `master` branch and smaller PRs, which get quicker rejected if they don't comply with the Project Guidelines.

Having a `develop` branch let's the developers sometimes commit real buggy code and then the core developers have to fix these issues again before merging it into `master`, which is just too time-consuming. 

The guys from Odoo are doing it the same (only one master branch where all PRs get in) and they have had a huge increase of people sending PRs to this project.

I hope we'll find some time soon with @rockfruit to discuss this into more detail, I would really appreciate it.

bika/lims/browser/analysisrequest/analysisrequests.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, context, request):
 
         self.catalog = "bika_catalog"
         self.contentFilter = {'portal_type': 'AnalysisRequest',
-                              'sort_on': 'created',
+                              'sort_on': 'Created',
```
rockfruit: This line should have been let as it was: the index in bika_catalog is named 'created' on bika/lims/setuphandlers.py:646

bika/lims/browser/analysisrequest/analysisrequests.py
```diff
@@ -696,8 +700,7 @@ def isItemAllowed(self, obj):
         if not self.context.bika_setup.getAllowDepartmentFiltering():
             return True
         # Gettin the department from analysis service
-        ans = [an.getObject() for an in obj.getAnalyses()]
-        deps = [an.getService().getDepartment().UID() for an in ans if an.getService().getDepartment()]
+        deps = obj.getDepartmentUIDs() if hasattr(obj, 'getDepartmentUIDs') else []
```
rockfruit: isItemAllowed in this context will only accept an AnalysisRequest; I can't explain why the test is here, but acquisition will not play a role.

It's confusing but it is not incorrect.

bika/lims/browser/analysisrequest/analysisrequests.py
```diff
@@ -718,22 +721,24 @@ def folderitem(self, obj, item, index):
...
             url += "/manage_results"
 
-        item['Client'] = obj.aq_parent.Title()
-        if (hideclientlink == False):
+        clientuid = obj.getClientUID()
+        client = self.clients.get(clientuid,None)
+        if not client:
+            client = obj.aq_parent
+            self.clients[clientuid]=client
+
```
rockfruit: Agree, using obj.getClient() will wake up the object, but subsequent calls in the same request will not affect performance notably.

bika/lims/browser/analysisrequest/analysisrequests.py
```diff
@@ -772,6 +776,21 @@ def folderitem(self, obj, item, index):
         item['getDateVerified'] = \
             self.ulocalized_time(getTransitionDate(obj, 'verify'))
 
+        if self.printwfenabled:
```
rockfruit: __call__ always uses super() or BaseClass.__call__ so subclass will not break.

Using __init__ is problematic for views, because current user and rules are not available there.

bika/lims/browser/analysisrequest/analysisrequests.py
```diff
@@ -824,8 +843,8 @@ def folderitem(self, obj, item, index):
             sampler = sample.getSampler().strip()
             if sampler:
                 item['replace']['getSampler'] = self.user_fullname(sampler)
-            if 'Sampler' in member.getRoles() and not sampler:
-                sampler = member.id
+            if 'Sampler' in self.member.getRoles() and not sampler:
```
rockfruit: As above; during __init__ member roles are not configured.

rockfruit:
I'm replying at @ramonski's comments, and I'll review the code myself afterwards and make some fixes.  I think closing the PR is drastic; sometimes we get in these situations and it's worth getting out using the quickest method that is easiest for everyone.

bika/lims/browser/analysisrequest/analysisrequests.py
```diff
@@ -897,6 +915,41 @@ def __call__(self):
...
+            #Print button to choose multiple ARs and print them.
+            review_states = []
+            for review_state in self.review_states:
+                review_state.get('custom_actions', []).extend(
+                    [{'id': 'print',
+                      'title': _('Print'),
+                      'url': 'workflow_action?action=print'}, ])
+                review_states.append(review_state)
+            self.review_states = review_states
+
```
rockfruit: Agree 100%

bika/lims/browser/analysisrequest/published_results.py
```diff
@@ -85,6 +87,22 @@ def __call__(self):
...
+            # Remove "Printed" columns
+            del self.columns[printed_colname]
+            tmprvs = []
+            for rs in self.review_states:
+                tmprs = rs
+                tmprs['columns'] = [c for c in rs.get('columns', []) if
+                                    c != printed_colname]
+                tmprvs.append(tmprs)
+            self.review_states = tmprvs
+
```
rockfruit: Agree, this is used in ~10-15 places through the code; but this refactoring should be a separate PR

bika/lims/browser/analysisrequest/workflow.py
```diff
@@ -450,6 +450,16 @@ def workflow_action_prepublish(self):
     def workflow_action_republish(self):
         self.workflow_action_publish()
 
+    def workflow_action_print(self):
+        # Calls printLastReport method for selected ARs
+        uids = self.request.get('uids',[])
+        uc = getToolByName(self.context, 'uid_catalog')
+        for obj in uc(UID=uids):
+            ar=obj.getObject()
```
rockfruit: uid_catalog is just a normal catalog although with fewer indexes; returns brains.

bika/lims/browser/widgets/referencewidget.py
```diff
@@ -138,6 +138,8 @@ def get_base_query(self, context, fieldName):
     def initial_uid_field_value(self, value):
         if type(value) in (list, tuple):
             ret = ",".join([v.UID() for v in value])
+        elif type(value) in [str, ]:
```
rockfruit: better is `isinstance(value, basestring)`

bika/lims/browser/widgets/reflexrulewidget.py
rockfruit: If getMethod() is called once in the request, zeoclient caches the result at least for the duration of the request; so multiple calls may be slower but not by much although I agree from a style point of view.

bika/lims/browser/widgets/serviceswidget.py
rockfruit: true;  getattr has a default parameter, so, this could be written as `if self.context.getattr('getRestrictToMethod', False):`

bika/lims/browser/analysisrequest/analysisrequests.py
rockfruit: The confusion here comes from the fact that the index is actually 'created' lowercase, but the name of the column in that file is 'Created'.  @ramonski added a logger.warn in bika_listing that alerts us to these conditions in future.

bika/lims/browser/analysisrequest/analysisrequests.py
rockfruit: I've removed the condition to reduce confusion here; if it fires an exception, we should fix in the calling code.  @xispa do you remember where that is?

bika/lims/content/analysisrequest.py
rockfruit: I am making sure in the listing view that the print button is only available in "published" filtered views, but still I will modify this to be more defensive.

bika/lims/browser/dashboard/aggregatedanalyses.py
```diff
@@ -0,0 +1,143 @@
...
+        """
+        It checks if the item can be added to the list depending on the
+        department filter. If the analysis service is not assigned to a
+        department, show it.
+        If department filtering is disabled in bika_setup, will return True.
+        @Obj: it is an analysis object.
+        @return: boolean
+        """
+        if not obj:
+            return None
```
rockfruit: well practically it does, because 'if isItemAllowed():' with None will be False :>  I changed it.

bika/lims/browser/department/ajax.py
```diff
@@ -0,0 +1,56 @@
...
+        # Getting the lab contact object
+        labcontact_brain = pc(
+            portal_type='LabContact', UID=dataset.get('contact_uid', ''))
+        # If the lab contact exists, update it
+        if labcontact_brain and dataset.get('checkbox_value', False):
+            # Getting the department
+            department = str(dataset.get('department'))
+            department = department[1:] if\
+                department.startswith('/') else department
+            dep = self.context.unrestrictedTraverse(department)
```
rockfruit: I don't know.  @xispa is there any reason that this should require traversal without security checks?  I think in this case, because we are not accepting user input, it does not make any difference.

rockfruit: I know in the past we have required the unrestrictedTraverse, and it took some time to debug to realise this is the reason for a failure; so now, I usually use unrestrictedTraverse by default, regardless of the context; and usually, we are in control of the code and we know that the required object is allowed to be accessed.

bika/lims/browser/fields/reflexrulefield.py
```diff
@@ -0,0 +1,295 @@
...
+        field info. This dictionaries must be in sync with the
+        browser/widgets/reflexrulewidget.py/process_form() dictionaries format.
+        """
+        for d in rules_list:
+            # Checking if all dictionary items are correct
+            if not _check_set_values(instance, d):
+                RecordsField.set(self, instance, [], **kwargs)
+        RecordsField.set(self, instance, rules_list, **kwargs)
+
+
```
rockfruit: I've moved them into the class because it feels as though they belong there, although they get no real benefit from the move.

rockfruit:
@xispa do you have any reason to require that this code is accepted into master before 3.2?  I would like to branch a curate/3.2 before accepting this PR, because maybe it would benefit from some time being tested in our test deployments before going into release?

bika/lims/browser/js/bika.lims.rejection.js
```diff
@@ -23,18 +23,13 @@
...
-                 // https://jira.bikalabs.com/browse/LIMS-2558
-                 if (rejection_reasons.length == 0) {
-                     console.debug("bika.lims.rejection.RejectionKickOff: No rejection reasons found in ", data);
-                     return;
-                 }
-                 var reasons_state = rejection_reasons[0]['checkbox'];
-                 if (reasons_state == undefined || reasons_state != 'on'){
+             if (data.success &&
+                 data.total_objects > 0) {
+                 var reasons_state = data.objects[0].RejectionReasons[0].checkbox;
```
rockfruit: SHould not; there is a check for `data.total_objects > 0` on line 26.

bika/lims/browser/js/bika.lims.rejection.js
```diff
@@ -48,7 +43,7 @@
          if($('div#archetypes-fieldname-RejectionWidget').length > 0){
              // binding a new click action to state's rejection button
              $("a#workflow-transition-reject").unbind();
-             $("a#workflow-transition-reject").click(function(e){
+             $("a#workflow-transition-reject").bind("click", function(e){
                  // Overlays the rejection widget when the user tryes to reject the ar and
```
rockfruit: @xispa fyi, `live` is also deprecated; jquery wants us to use `on` from now.  So a simple refactory will happen in future for this purpose

bika/lims/browser/widgets/referencewidget.py
rockfruit: It seems for the check on line 139 above, there is no equally simple check for `basesequence` to cover both list and tuple.

bika/lims/browser/widgets/reflexrulewidget.py
rockfruit: on second look, as a style issue mostly, I changed this.

bika/lims/browser/widgets/serviceswidget.py
rockfruit: @pau I don't understand why hasattr() checks getRestrictToMethod, but getattr() uses getMethodUID?  It seems incorrect, or at least it's confusing.

bika/lims/content/analysisrequest.py
rockfruit: I'm not sure that this is "pythonic" but I agree more readable, I changed it.

xispa:
@rockfruit ,  I know you have no shortage of things to do, so many thanks for reviewing. I don't have any strong reason for this code to be accepted on 3.2, except this one includes changes for some PR accepted earlier. E.g.: #1866 #1868 #1869 etc..

bika/lims/browser/analysisrequest/analysisrequests.py
xispa: Since you are listing ARs here and some ARs will have the same client, I thought there wasn't the need to retrieve the client for each AR, rather only from ARs the Client wasn't retrieved yet.

bika/lims/content/bikacache.py
xispa: Will ask @Nihadness . This was used as a way to make the `nextid` thing (in idserver.py) much faster, cause at that moment, the `nextid` function was walking through all objects to just only get the ID prefix from the latest object. In our current code we dismissed this approach and we are using a direct query to the catalog instead.

bika/lims/browser/analysisrequest/workflow.py
ramonski: Yes you're right. As long as you don't use `uc.lookupObject(uid)`, you get catalog brains.

ramonski: ... but maybe I'm confusing this with the `reference_catalog` 🤓

xispa:
@ramonski I don't think the Odoo's approach work in our case. I guess they have a very large community of devs, so they can afford the push some of them (and their developments) in a "stale" situation. This is something I feel we cannot afford. IMO, the same would apply clients, not only to devs.

Re overhead, I think having only master doesn't mean you have less overhead than using master+develop+support. Also if it generates less overhead, the problem is that using master (only) approach, you are forced to address the overhead while you are in the middle of an implementation project for a given client, with a tight timeline and deliverables. Having master+develop+support would allow us to apply the "overhead", at least, outside of a critical implementation process.

Re Project Guidelines, totally agree with you. PRs should follow strict rules, either if the target branch is master or develop or support

ramonski:
@xispa I am having still a strong different opinion about this.

I think it is not a good approach to have a customer project integrated on a GitHub branch at all.
So I think the release cycles are the actual problem here.

I think it is never possible to overlook the side-effects of a PR, and having it on `master`, it will at least ensure that we will realize it quicker than through the indirection of a `develop/support` branch, which makes things more complicated in my opinion.

We need to get more snappy, both with the changes in code and releases of `bika.lims` to attract more developers and of course Customers, and I think that is something that the Odoo guys have done right.

So I would recommend to be more picky for PRs which go into `master` to get the right quality and grew a developer mentality of a "single PRs for a single purposes".

Furthermore, I would like to have more releases out of `master` (maybe once a month?), which would eliminate the need to have customer projects run on a branch.

As I said, this is just my point of view, but does this approach make sense to you anyhow?

xispa:
@ramonski your approach makes totally sense to me, and I wouldn't have any problem to adopt it if there was developers destined to the base code only, regardless of client-specific projects. But the truth is (at least, until now) that we've been fixing and improving the code based on the needs of clients, so we all have been contributing to the code almost at the same pace as the clients needs. 

Your approach assumes we must charge the overhead of dealing with fast-releases (1month) to our clients, both in terms of work and costs. Take into account that most of the projects take at least 3 months time, which means that, apart from the functionalities/fixes/improvements/customizations the client-specific project require, we'll need to accommodate time for upgrading their code every month, which  IMO is not feasible because our development sprints take 4 weeks to complete plus one additional week for acceptance testing, each one. Also, note that we also create project-specific add-ons for clients and we must ensure they are compatible with the current Bika LIMS (or Bika Health, don't forget it, please!) version. Believe me, I'd like to be wrong, but I feel this would definitely spoil the work, costs and timeline we agreed with clients.... and remember, they are the payers.

Re single PR for single-purpose, totally agree.

ramonski:
@xispa, @rockfruit I would suggest that we do a Google-Hangout to coordinate how we continue with the development process, maybe tonight 6 pm?

As I said xispa, as a fresh member of the Bika family, I did not meant to drop existing and evolved processes. I understand your point of view completely.

I'm just coming from another background and have therefore a different view on how to do software projects.

Anyhow, I would recommend that we should adapt our customer projects to the `bika.lims` process and not vice versa.

If we built a customer extension for e.g. `bika.lims=3.2.1` the customer has to pay for an upgrade to `bika.lims=3.2.2` and so on.
It can not be that our customers dictate how we do proper software engineering, just to keep their costs low!

Building releases can be furthermore pretty much automated and gives us for every version a clean changelog and small upgrade steps. 

But let us talk about this personally, I would appreciate it.

Processes are there to get optimized. I'm sure we find the right one where we can all adapt to.

xispa:
@ramonski @rockfruit I was thinking that would be better if we discuss this in the devs list.

lemoene:
Phew that give me a headache...

I made repo management part of the Bika 4 scope too, that project definituon to cycle for review real soon now.

From a mile up, I'll put my weight behind your decision, if it would

- keep a slim trunk with 'extensions' for clients they pay the maintenance of

- facilitate modular add-ons (in future), e.g. Sampling Rounds, Invoices, etc. that are not required by all

- keep regression errors in check

- make for shorter release cycles

When you agree on how it should be done, please document in the wiki and start 'migration'-)

------------------------------------

Is reflex testing going into Health only? If not directly in LIMS, it'd be a good use case for making it an add-on (when a client asks and sponsors)

xispa:
@lemoene reflex testing goes to Bika LIMS, but this functionality hasn't been coded as an add-on.

> keep a slim trunk with 'extensions' for clients they pay the maintenance of

AFAK, we do this already. We create project-specific add-ons that only contain those things that we don't need in the official extension (Bika LIMS or Bika Health). 

> facilitate modular add-ons (in future), e.g. Sampling Rounds, Invoices, etc. that are not required by all

This would be ideal, but keep in mind that this means an additional (and imo, not negligible) overhead: each add-on needs to be maintained by someone, at least to guarantee its compatibility with the latest version of the Bika LIMS "core". You can take Bika Health (or any project-specific add-on) as an example: we have to allocate time for revisiting the add-on each time we release a new version of Bika LIMS. Although I think a modular approach is an ideal solution, I have the feeling Plone will make this approach much harder than we initially would expect - we could end-up with a bunch of different add-ons with strong dependencies amongst them. IMHO, changing to the modularity paradigm will be an opportunity to embrace a new framework. And if the decision is to adopt a new framework, then I'd say that we should take advantage of that situation for revisiting the basis of the project itself (migrate to a postgreSQL, adopt a real MVC architecture, write a Code of Conduct, write a Document of Done, etc.). In summary, it would be like almost starting a project from scratch.

> keep regression errors in check

Yes, @ramonski  suggestions regarding Guidelines and PRs strict rules go in this direction. And I am totally agree with this. But we need a way to also ensure the new contributions guarantee backwards (and derivative add-ons) compatibility.

> make for shorter release cycles

@ramonski was suggesting a 1m release-cycle, my opinion about this here: https://github.com/bikalabs/bika.lims/pull/1928#issuecomment-281056474 
I still think a release every 6m is a good compromise between a community-driven-approach and a project-driven-approach. Even better for support branches: since they would only include fixes and typos, I feel we could deploy a new release every <=3m.

lemoene:
Not sure whether mailing github works..
>
>     facilitate modular add-ons (in future), e.g. Sampling Rounds,
>     Invoices, etc. that are not required by all
>
> This would be ideal, but keep in mind that this means an additional 
> (and imo, not negligible) overhead: each add-on needs to be maintained 
> by someone, at least to guarantee its compatibility with the latest 
> version of the Bika LIMS "core".
>

Clients will simply have to own up. There is no way around it, unless it 
can be coded in such a way that if disabled, the function will be out of 
sight an have no impact on performance. sticky

> You can take Bika Health (or any project-specific add-on) as an 
> example: we have to allocate time for revisiting the add-on each time 
> we release a new version of Bika LIMS. Although I think a modular 
> approach is an ideal solution, I have the feeling Plone will make this 
> approach much harder than we initially would expect - we could end-up 
> with a bunch of different add-ons with strong dependencies amongst 
> them. IMHO, changing to the modularity paradigm will be an opportunity 
> to embrace a new framework. And if the decision is to adopt a new 
> framework, then I'd say that we should take advantage of that 
> situation for revisiting the basis of the project itself (migrate to a 
> postgreSQL, adopt a real MVC architecture, write a Code of Conduct, 
> write a Document of Done, etc.). In summary, it would be like almost 
> starting a project from scratch.
>

All for Bika four. See project definition circulated. Can't wait to get 
to this level.

>     keep regression errors in check
>
> Yes, @ramonski <https://github.com/ramonski> suggestions regarding 
> Guidelines and PRs strict rules go in this direction. And I am totally 
> agree with this. But we need a way to also ensure the new 
> contributions guarantee backwards (and derivative add-ons) compatibility.
>
>     make for shorter release cycles
>
> @ramonski <https://github.com/ramonski> was suggesting a 1m 
> release-cycle, my opinion about this here: #1928 (comment) 
> <https://github.com/bikalabs/bika.lims/pull/1928#issuecomment-281056474>
> I still think a release every 6m is a good compromise between a 
> community-driven-approach and a project-driven-approach. Even better 
> for support branches: since they would only include fixes and typos, I 
> feel we could deploy a new release every <=3m.
>

Once we are in a new framework  we could revisit this.

xispa:
>  Clients will simply have to own up. [..]

Clients will simply dismiss this offer imo.

> [..] unless it can be coded in such a way that if disabled, the function will be out of sight an have no impact on performance. sticky

Having a lot of functionalities in the same code don't necessarily mean a significant impact in performance, rather increases the code complexity, implies less flexibility and makes the app less reusable. In exchange, maintaining 1 product is easier than maintaining 20 products. This being said, I still recognize that multiple add-ons might be a better approach in a long-term run. It will mainly depend on being able to keep a reasonable number of add-ons, with enough maintainers, resources and clients ensuring their viability.

> All for Bika four. See project definition circulated. Can't wait to get to this level.

Meanwhile, we are still in Bika 3, and lot of implementations rely on 3.

> Once we are in a new framework we could revisit this.

Ummm... and until then, what do you suggest?

lemoene:
>     Clients will simply have to own up. [..]
>
> Clients will simply dismiss this offer imo.
>

We manoeuvred a few in that direction already. Prospects still have a 
choice, but all work be billable

>     [..] unless it can be coded in such a way that if disabled, the
>     function will be out of sight an have no impact on performance. sticky
>
> Having a lot of functionalities in the same code don't necessarily 
> mean a significant impact in performance,
>

Has to be guarded against, i imagine redundant indexing may take place

> rather increases the code complexity, implies less flexibility and 
> makes the app less reusable.
>

Indeed, and if not hidden properly - from that perspective Invoices and 
Partitions were done relatively well - a cluttered UI

> In exchange, maintaining 1 product is easier than maintaining 20 
> products.
>

The client's prerogative to decide

In many cases, it will be small differentiations, e.g. Bika Wine for the 
beverages sector, and Bika Cannabis

> This being said, I still recognize that multiple add-ons might be a 
> better approach in a long-term run. It will mainly depend on being 
> able to keep a reasonable number of add-ons, with enough maintainers, 
> resources and clients ensuring their viability.
>

jip

>     All for Bika four. See project definition circulated. Can't wait
>     to get to this level.
>
> Meanwhile, we are still in Bika 3, and lot of implementations rely on 3.
>

The Bika 4 approach to be decided soonest and new code developed with 
that in mind.

One of the Bika 4 objectives should be to re-use as much code possible. 
Not necessarily in the code and wrt to functional modularity, but we 
should consider as intermediary steps making use of Bika 3 and Newt, 
standard JS, and other platform independent stuff we can use in Bika 4 
too or phase out there.

I'll start draft documents for the framework evaluation and functional 
spec soon, I think everybody is primed for discussing the former, we can 
move it to the lists yes for wider input and the benefit of the Google 
crawler. First we need to nail the project definition. Not difficult, 
see draft. Probably a bit 90's, should still serve its purpose. 
@ramonski's how do the business schools do it nowadays?

I need to focus on our grant application. Please see what is available 
in Spain, Germany EU, and we do a South-North project...

ramonski:
I’ll answer short, as I would as well suggest to have that discussion continued
in the dev-lists and also they will switch off the electricity here in a couple of minutes;)

> > Clients will simply have to own up. [..]
> >
> > Clients will simply dismiss this offer imo.
> >
> 
> We manoeuvred a few in that direction already. Prospects still have a 
> choice, but all work be billable
> 
> > [..] unless it can be coded in such a way that if disabled, the
> > function will be out of sight an have no impact on performance. sticky
> >
> > Having a lot of functionalities in the same code don't necessarily 
> > mean a significant impact in performance,
> >
> 
> Has to be guarded against, i imagine redundant indexing may take place
> > rather increases the code complexity, implies less flexibility and 
> > makes the app less reusable.
> >
> 
> Indeed, and if not hidden properly - from that perspective Invoices and 
> Partitions were done relatively well - a cluttered UI
> 

Indeed, a monolytic approach doesn’t automatically mean, that the application
gets slower, but it is harder to maintain and it is easier to get side effects somewhere
if one changes the functionality on one end.

> > In exchange, maintaining 1 product is easier than maintaining 20 
> > products.
> >
> 
> The client's prerogative to decide
> 
> In many cases, it will be small differentiations, e.g. Bika Wine for the 
> beverages sector, and Bika Cannabis
> 
> > This being said, I still recognize that multiple add-ons might be a 
> > better approach in a long-term run. It will mainly depend on being 
> > able to keep a reasonable number of add-ons, with enough maintainers, 
> > resources and clients ensuring their viability.
> >
> 
> jip
> 
> > All for Bika four. See project definition circulated. Can't wait
> > to get to this level.
> >
> > Meanwhile, we are still in Bika 3, and lot of implementations rely on 3.
> >
> 
> The Bika 4 approach to be decided soonest and new code developed with 
> that in mind.
> 
> One of the Bika 4 objectives should be to re-use as much code possible. 
> Not necessarily in the code and wrt to functional modularity, but we 
> should consider as intermediary steps making use of Bika 3 and Newt, 
> standard JS, and other platform independent stuff we can use in Bika 4 
> too or phase out there.
> 
> I'll start draft documents for the framework evaluation and functional 
> spec soon, I think everybody is primed for discussing the former, we can 
> move it to the lists yes for wider input and the benefit of the Google 
> crawler. First we need to nail the project definition. Not difficult, 
> see draft. Probably a bit 90's, should still serve its purpose. 
> @ramonski's how do the business schools do it nowadays?

In this stage we need some kind of direction where we all agree on.
So I think your project definition document has still its justification and
we should go on with this on a high level perspective.

Open source projects are special and hard to press into a conventional
project management processes or even agile PM, as most of the people
contribute w/o getting paid and thus, we simply don’t have at the moment
the necessary resources I guess.

Anyhow, here are my 2 cents about the further steps:

- Definition of an „ideal“ core functionality

- Definition of the process flows for this functionality

- Make the whole Bika Core kind of headless and develop a JS Application which
  communicates with it through JSON, similar to the search view I showed
  recently (not standalone like Bika Commander).

> I need to focus on our grant application. Please see what is available 
> in Spain, Germany EU, and we do a South-North project...

ramonski:

ramonski:

PR #1930, opened 2017-02-10
Fix LIMS-1807 (PR: #1757)

rockfruit:


bika/lims/validators.py
```diff
@@ -120,15 +120,11 @@ class InvoiceBatch_EndDate_Validator:
...
-        # form = request.get('form', {})
-        enddate = value
-        startdate = startdate.strftime('%Y-%m-%d %H:%M')
-
         translate = getToolByName(instance, 'translation_service').translate
 
-        if not enddate >= startdate:
+        startdate = kwargs.get('REQUEST').get('BatchStartDate', '')
+        enddate = kwargs.get('REQUEST').get('BatchEndDate', '')
+        if enddate <= startdate:
```
ramonski: Convert `startdate` and `enddate` to `DateTime` objects, otherwise it will be a string comparison, which will probably give funny results for german datetime formats (`17.02.2017`)

ramonski:
I think this one is just wrong and shouldn't be merged (see comment). However, should be simple to fix...

PR #1939, opened 2017-02-14
LIMS-2500 - Columns to be added to Sample Point lists

Lunga001:

https://jira.bikalabs.com/browse/LIMS-2500

## Current behavior before PR
There was Title and Description columns only

Added other columns Sample Types, Composite Y/N, Sampling frequency, Attachments (hyperlinked)
And using folderitem instead of folderitems

--

### DIFF
--- a/bika/lims/browser/client/views/samplepoints.py
+++ b/bika/lims/browser/client/views/samplepoints.py
@@ -39,22 +39,30 @@ class ClientSamplePointsView(BikaListingView):
                       'replace_url': 'absolute_url'},
             'Description': {'title': _('Description'),
                             'index': 'description'},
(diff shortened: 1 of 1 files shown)


bika/lims/browser/client/views/samplepoints.py
```diff
@@ -65,3 +73,30 @@ def __call__(self):
                 {'url': 'createObject?type_name=SamplePoint',
                  'icon': '++resource++bika.lims.images/add.png'}
         return super(ClientSamplePointsView, self).__call__()
+
+    def folderitems(self):
```
ramonski: Instead of looping the items in `folderitems`, you should use the `folderitem` method. This gets called for every item, which simplifies things.

PR #1941, opened 2017-02-15
LIMS-2509-Fixed values that were not saving when cleared

Lunga001:

https://jira.bikalabs.com/browse/LIMS-2509

## Current behavior before PR
Fields were not clearing after being edited
New values are saved after being edited

--


bika/lims/browser/widgets/coordinatewidget.py
```diff
@@ -4,17 +4,33 @@
 # Some rights reserved. See LICENSE.txt, AUTHORS.txt.
 
 from AccessControl import ClassSecurityInfo
-from Products.ATExtensions.widget import RecordWidget
+from Products.Archetypes.Widget import TypesWidget
 from Products.Archetypes.Registry import registerWidget
 import datetime
 
-class CoordinateWidget(RecordWidget):
+class CoordinateWidget(TypesWidget):
```
ramonski: Why is it needed to change the base class of this Widget?

bika/lims/browser/widgets/coordinatewidget.py
```diff
@@ -4,17 +4,33 @@
...
-    _properties = RecordWidget._properties.copy()
+    _properties = TypesWidget._properties.copy()
     _properties.update({
         'macro': "bika_widgets/coordinatewidget",
     })
 
+    def process_form(self, instance, field, form, empty_marker=None,
+                     emptyReturnsMarker=False, validating=True):
+        """Basic impl for form processing in a widget"""
+        empty_marker = {} #This is a hack
```
ramonski: indeed. This should not go into `master`

rockfruit:
There was some very strange behaviour happening here; the BaseObject.py processForm (which calls into the CoordinateWidget/process_form function) was setting empty_marker to [].  In this case however the empty value for these fields is always {}.

PR #1943, opened 2017-02-15
Feature/instrument interface linkage

xispa:
## Current behavior before PR
In the Instrument Import, the user has to select the interface to use and at the same time, has to select the instrument (in a selection list at the bottom of the form) to which the import refers, so the system will be able then to assign the result to a registered instrument and also find out if the analysis is currently an internal calibration test.

An "Import Interfaces" field has been added in the Instrument's edit view, so the user can choose which are the interfaces that can be used for importing results generated by that instrument. Note the field allows to assign more than one interface to an instrument.

In the Instrument Import view, the selection list at the bottom has been moved at the top of the form, and once the user selects the instrument, a second selection list for choosing one of the interfaces assigned to that selected instrument will appear. 

As a result, only the interfaces that at least have one instrument assigned will be available to the user for its selection, regardless of the interfaces available by default in Bika LIMS.

--


bika/lims/browser/js/bika.lims.instrument.import.js
```diff
@@ -10,6 +10,35 @@ function InstrumentImportView() {
...
+                       'instrument_uid': $(this).val()
+                      },
+                success: function(data){
+                  $("#exim").empty();
+                  $('#exim').append($('<option>').text('Choose an Interface...')
+                            .attr('value', '').attr('selected', 'selected'));
+                  $('#exim').append($('<option>').text('Default')
+                            .attr('value', 'default'));
+                  $.each(data, function(i, value) {
+                     $('#exim').append($('<option>').text(value.title).attr('value', value.id));
```
ramonski: I would suggest to add a function to do this, e.g.:
```
function add_option(id, text, value) {
    return $(id).append($('<option>').text(text).attr('value', value)
}
```

The returned element can then be chained to set the conditional "selected" attribute.

Being picky now;)

bika/lims/content/instrument.py
```diff
@@ -171,16 +172,26 @@
...
-        required_subfields=('Key', 'Value'),
-        subfield_labels={
-            'OptionValue': _('Key'),
-            'OptionText': _('Value'),
-        },
-        widget=RecordsWidget(
+    StringField('ImportDataInterface',
+        vocabulary = "getImportDataInterfacesList",
+        multiValued=1,
+        widget = MultiSelectionWidget(
```
ramonski: The `MultiSelectionWidget` import is missing.

bika/lims/content/instrument.py
```diff
@@ -340,6 +351,20 @@ def getDataInterfaces(context, export_only=False):
     exims.insert(0, ('', t(_('None'))))
```
ramonski: I am always confused about the magic `t` function. Would this value not be translated if we would just insert `_('None')`?

bika/lims/exportimport/dataimport.py
```diff
@@ -85,6 +86,13 @@ def __call__(self):
         else:
             return self.template()
 
+    def getInstruments(self):
+        bsc = getToolByName(self, 'bika_setup_catalog')
+        items = [('', '...Choose an Instrument...')] + [(o.UID, o.Title) for o in
```
ramonski: Missing translation

bika/lims/exportimport/dataimport.py
```diff
@@ -85,6 +86,13 @@ def __call__(self):
...
+                               bsc(portal_type = 'Instrument',
+                                   inactive_state = 'active')]
+        items.sort(lambda x, y: cmp(x[1].lower(), y[1].lower()))
+        return DisplayList(list(items))
```
ramonski: don't we have already a list?

bika/lims/exportimport/dataimport.py
```diff
@@ -122,3 +122,32 @@ def getAnalysisServicesDisplayList(self):
...
+    def __call__(self):
+        interfaces = []
+        try:
+            plone.protect.CheckAuthenticator(self.request)
+        except Forbidden:
+            return json.dumps(interfaces)
+
+        from bika.lims.exportimport import instruments
+        bsc = getToolByName(self, 'bika_setup_catalog')
+        instrument=bsc(portal_type='Instrument',
```
ramonski: rename to `instruments`

bika/lims/exportimport/dataimport.py
```diff
@@ -122,3 +122,32 @@ def getAnalysisServicesDisplayList(self):
...
+                       UID=self.request.get('instrument_uid', ''),
+                       inactive_state='active',)
+        if instrument and len(instrument) == 1:
```
ramonski: This swallows the case if `len(instruments)>1`, which means that the `UID` catalog is broken. We should fail then.

bika/lims/exportimport/dataimport.py
```diff
@@ -122,3 +122,32 @@ def getAnalysisServicesDisplayList(self):
...
+            instrument = instrument[0].getObject()
+            for i in instrument.getImportDataInterface():
+                if i:
```
ramonski: Change variable name and reduce complexity (level of indentation).
Better use:
```
for interface in instrument.getImportDataInterface():
    if not interface:
        continue
    [...]
```

bika/lims/exportimport/import.pt
```diff
@@ -28,14 +28,26 @@
 
         <fieldset id="fieldset-instrumentimport">
             <legend id="fieldsetlegend-instrumentimport" i18n:translate="">Instrument Import</legend>
-            <select name="exim" id="exim"
-                tal:define="importers view/getDataInterfaces">
-                <tal:options repeat="option importers">
+            <br>
```
ramonski: we should avoid explicit line breaks.

bika/lims/exportimport/import.pt
```diff
@@ -28,14 +28,26 @@
...
+            <label for="instrument">Instrument</label><br>
+            <select name="instrument" id="instrument_select"
+                tal:define="instruments view/getInstruments">
```
ramonski: change loop variable to `instrument`

bika/lims/exportimport/import.pt
```diff
@@ -28,14 +28,26 @@
...
+                <tal:options repeat="option instruments">
                     <option
                         tal:attributes="value python:option;"
-                        tal:content="python:importers.getValue(option)"/>
+                        tal:content="python:instruments.getValue(option)"/>
                 </tal:options>
             </select>
+            <p i18n:translate="" style='color: #3F3F3F;font-size: 0.87em;'>
```
ramonski: Avoid to use explicit `style` attributes. Use CSS Classes instead.

bika/lims/exportimport/import.pt
```diff
@@ -28,14 +28,26 @@
...
+                If the system doesn't find any match (AnalysisRequest, Sample, Reference Analysis or Duplicate),
+                it will use the record's identifier to find matches with Reference Sample IDs.
+                If a Reference Sample ID is found, the system will automatically create a
+                Calibration Test (Reference Analysis) and will link it to the instrument selected above.<br/>
+                If no instrument selected, no Calibration Test will be created for orphan IDs.
+            </p>
+            <label for="instrument">Import Interface</label><br>
```
ramonski: - Missing `i18n:translate`
- Avoid explicit `<br/>`

ramonski:

PR #1871, opened 2016-11-24
LIMS-2321: Allow multiple Methods per Instrument

ramonski:
## Description

Allow multiple Methods per Instrument
https://jira.bikalabs.com/browse/LIMS-2321

**Please note**
This PR introduces 2 new helper modules, which introduce functional helpers for the well-being of all Bika developers.

- `bika.lims.api` introduces framework helpers which interfaces with Plone/Bika (One function for a sinle purpose)
- `bika.lims.utils.functools` introduces functional Python helpers (One function for a single purpose)

Both modules are easy to extend and help (at least me) to write less and do more;)

PR #1944, opened 2017-02-16
Feature/multi method per instrument

xispa:
## Current behavior before PR

Although more than one method can be assigned to an analysis and more than one instrument can be assigned to one method, is not possible to assign more than one method to the same instrument.

In the Instrument's edit view, the user can assign more than one method to that instrument through a multi-selection widget.

Some codes from #1871 also were used.

--


bika/lims/browser/instrument.py
```diff
@@ -620,24 +620,37 @@ def __init__(self, context, request):
...
         bsc = getToolByName(self, 'bika_setup_catalog')
-        instrument = bsc(portal_type='Instrument', UID=self.request.get("uid", '0'))
-        if instrument and len(instrument) == 1:
-            method = instrument[0].getObject().getMethod()
-            if method:
-                methoddict = {'uid': method.UID(),
-                              'title': method.Title()}
-        return json.dumps(methoddict)
+        results = bsc(portal_type='Instrument', UID=self.request.get("uid", '0'))
+        instrument = results[0] if results and len(results) == 1 else None
```
ramonski: I have seen this pattern quite a lot in the code and find it kind of cryptic.

We should better use a more pythonic approach like this:
```
results = bsc(portal_type='Instrument', UID=self.request.get("uid", '0'))
if results:
    instrument = results[0]
```

Also this code swallows an important exception, when we get _more_ than one result for an instrument UID. Would this case happen, then this function would return nothing.

So if this is really needed, we should fail:
```
if len(results) > 1:
    raise RuntimeError("More than one instrument found for UID %s" % self.request.get("uid", '0'))
```

bika/lims/content/analysisservice.py
```diff
@@ -1211,10 +1209,10 @@ def getAvailableMethods(self):
             # Add the methods from the instruments capable to perform
             # this analysis service
             for ins in self.getInstruments():
-                method = ins.getMethod()
-                if method and method.UID() not in muids:
-                    methods.append(method)
-                    muids.append(method.UID())
+                for method in ins.getMethods():
+                    if method and method.UID() not in muids:
```
ramonski: Maybe a dictionary would be better here, e,g, `{method.UID(): method}` and return dict.values() at the end.

bika/lims/content/instrument.py
```diff
@@ -391,6 +404,17 @@ def getManufacturers(self):
         items.sort(lambda x, y: cmp(x[1], y[1]))
         return DisplayList(items)
 
+    from bika.lims import deprecated
```
ramonski: This should go to the top of the file

bika/lims/exportimport/setupdata/__init__.py
```diff
@@ -772,7 +772,7 @@ def Import(self):
             obj.setInstrumentType(instrumenttype)
             obj.setManufacturer(manufacturer)
             obj.setSupplier(supplier)
-            obj.setMethod(method)
+            obj.setMethod([method])
```
ramonski: good catch

bika/lims/upgrade/to320.py
```diff
@@ -186,15 +190,6 @@ def create_CAS_IdentifierType(portal):
...
     """
-    pc = getToolByName(portal, 'portal_catalog', None)
-    objs = pc(portal_type="Analyses",review_state="to_be_verified")
-    for obj_brain in objs:
-        obj = obj_brain.getObject()
-        old_field = obj.Schema().get("NumberOfVerifications", None)
-        if old_field:
-            new_value=''
-            for n in range(0,old_field):
-                new_value+='admin'
```
ramonski: This change seems to not belong into this PR

bika/lims/upgrade/to320.py
```diff
@@ -205,21 +200,39 @@ def create_CAS_IdentifierType(portal):
...
     Getting all analyses with review_state in to_be_verified and
     adding "admin" as a verificator as many times as this analysis verified before.
     """
     pc = getToolByName(portal, 'portal_catalog', None)
-    objs = pc(portal_type="Analyses",review_state="to_be_verified")
+    objs = pc(portal_type="Analyses", review_state="to_be_verified")
     for obj_brain in objs:
         obj = obj_brain.getObject()
-        old_field = obj.Schema().get("NumberOfVerifications", None)
+        old_field = obj.Schema().get("NumberOfVerifications", None).get(obj)
```
ramonski: better use `obj.getField("NumberOfVerifications")`

bika/lims/browser/instrument.py
rockfruit: perhaps simply `assert len(results) < 2`

rockfruit: Modified to use uid_catalog - when using uid_catalog everyone (including all Plone code I browsed) assumes that UIDs are unique if present.

bika/lims/content/analysisservice.py
rockfruit: done

bika/lims/upgrade/to320.py
rockfruit: meh, it's an upgrade step, and a good fix, I'll leave it

rockfruit:
Bansai!

bika/lims/content/analysisservice.py
```diff
@@ -1186,14 +1186,12 @@ def getMethod(self):
...
+        # is correct.
         method = None
-        if (self.getInstrumentEntryOfResults() == True):
-            method = self.getInstrument().getMethod() \
-                if (self.getInstrument() \
-                    and self.getInstrument().getMethod()) \
-                else None
-        else:
-            method = self.get_Method();
+        if (self.getInstrumentEntryOfResults() is True):
```
rockfruit: If `Allow manual entry of results` is on, then we can select a `DefaultMethod` (_Method) from methods selected manually.

If `Allow instrument entry of results` is on, then we can select a `DefaultMethod` from a group of methods which are provided by selected instruments.

Why then, does the line above only return the _Method field (`DefaultMethod`) if the `InstrumentEntry` field is True?

xispa: I see @rockfruit , I'd say is better to return the first method from manual entry if the instrument entry of results is off:

    method = None
    if self.getInstrumentEntryOfResults():
        method = self.get_Method()
    else:
        method = self.getMethods()
        method = method[0] if method else None
    return method

PR #1945, opened 2017-02-16
LIMS-2333 Added to context/path to search query on ajax call

Lunga001:
All Samples are listed for a client should only list its own
https://jira.bikalabs.com/browse/LIMS-2333

## Current behavior before PR
    On AR Add samples: all samples for all clients are shown 
    Should only list samples that belong to the current client
--

### DIFF
--- a/bika/lims/browser/widgets/referencewidget.py
+++ b/bika/lims/browser/widgets/referencewidget.py
@@ -192,10 +192,6 @@ class ajaxReferenceWidgetSearch(BrowserView):
         rows = []
 
         brains = []
-        self.request['search_query'] = json.dumps(
-                {'path' : {"query": 
-                            '/'.join(self.context.getPhysicalPath()) }
-                })
         for name, adapter in getAdapters((self.context, self.request), IReferenceWidgetVocabulary):
             brains.extend(adapter())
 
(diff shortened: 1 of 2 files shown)


bika/lims/browser/widgets/referencewidget.py
```diff
@@ -192,6 +192,10 @@ def __call__(self):
         rows = []
 
         brains = []
+        self.request['search_query'] = json.dumps(
+                {'path' : {"query": 
+                            '/'.join(self.context.getPhysicalPath()) }
+                })
```
ramonski: Please describe in your PR why you're doing this.

rockfruit: https://github.com/bikalabs/bika.lims/blob/master/bika/lims/browser/js/bika.lims.analysisrequest.add_by_col.js#L391 - there is some code to filter the field lookups via javascript when they are rendered.

At first I thought this should be done purely with permissions, but as one user can have 'view' permission on samples from multiple clients, this needs to be done in the JS regardless.

With the current solution above, it seems that the path-query will always be inserted into the request, even in some cases where it's not wanted.

Lunga001:
Removed the changes on server-side and added it on js

PR #1947, opened 2017-02-16
Bika LIMS API

ramonski:

The Bika LIMS API makes your live immediately more pleasant!

It provides _single_ functions for _single_ purposes, it is tested, it is mighty, it is GREAT!

## Current behavior before PR

Common tasks like getting a tool, searching the catalog, checking the workflow state etc., were always repeated, or non-efficient functions were used, e.g. `getToolByName`.

The API acts as a central interface for common tasks done in Bika LIMS and follows the principle of DRY (Don't Repeat Yourself).

This will make Bika LIMS perform better, supports more readable and easier maintainable code and it avoids errors by non-core developers.

Please read the Doctest in `bika.lims/docs/API.rst` for further details.

--

### DIFF
--- a/bika/lims/api.py
+++ b/bika/lims/api.py
@@ -20,9 +20,25 @@ from plone.api.exc import InvalidParameterError
 from plone.dexterity.interfaces import IDexterityContent
 from plone.app.layout.viewlets.content import ContentHistoryView
 
-# from bika.lims import logger
+"""Bika LIMS Framework API
 
-"""Please see bika.lims/docs/API.rst for documentation
+Please see bika.lims/docs/API.rst for documentation.
+
+Architecural Notes:
+
+Please add only functions that do a single thing for a single object.
+
+Good: `def get_foo(brain_or_object)`
+Bad:  `def get_foos(list_of_brain_objects)`
+
+Why?
+
+Because it makes things more complex. You can always use a pattern like this to
+achieve the same::
+
+    >>> foos = map(get_foo, list_of_brains_objects)
+
+Please add for all of your functions a descriptive test in docs/API.rst. Thanks.
 """
 
 _marker = object()
@@ -60,8 +76,7 @@ def create(container, portal_type, title=None, **kwargs):
     :returns: The new created object
     :rtype: object
     """
-
-    title = title is None and "Test {}".format(portal_type) or title
+    title = title is None and "New {}".format(portal_type) or title
     _ = container.invokeFactory(portal_type, id="tmpID", title=title)
     obj = container.get(_)
     obj.processForm()
--- a/bika/lims/api.py
+++ b/bika/lims/api.py
@@ -36,7 +36,7 @@ Why?
 Because it makes things more complex. You can always use a pattern like this to
 achieve the same::
 
-    >>> foos = map(get_foo, list_of_brains_objects)
+    >>> foos = map(get_foo, list_of_brain_objects)
 
 Please add for all of your functions a descriptive test in docs/API.rst. Thanks.
 """
--- a/bika/lims/api.py
+++ b/bika/lims/api.py
@@ -224,6 +224,19 @@ def get_id(brain_or_object):
     return get_object(brain_or_object).getId()
 
 
+def get_title(brain_or_object):
+    """Get the Title for this object
+
+    :param brain_or_object: A single catalog brain or content object
+    :type brain_or_object: ATContentType/DexterityContentType/CatalogBrain
+    :returns: Title
+    :rtype: string
+    """
+    if is_brain(brain_or_object):
+        return brain_or_object.Title
+    return get_object(brain_or_object).Title()
+
+
 def get_uid(brain_or_object):
     """Get the Plone UID for this object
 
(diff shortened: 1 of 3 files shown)


bika/lims/api.py
```diff
@@ -0,0 +1,774 @@
...
+
+    :param path: The physical path of the object to find
+    :type path: string
+    :returns: Found Object or None
+    :rtype: object
+    """
+
+    # nothing to do here
+    if not path:
+        return None
```
rockfruit: I think we should raise here, so that get_object_by_path (if it returns) __always__ returns an object.

bika/lims/api.py
```diff
@@ -0,0 +1,774 @@
...
+
+    if not path.startswith(portal_path):
+        fail("Not a physical path inside the portal.")
+
+    if path == portal_path:
+        return portal
+
+    res = pc(path=dict(query=path, depth=0))
+    if not res:
+        return None
```
rockfruit: here too, I think we should fail.

bika/lims/api.py
```diff
@@ -0,0 +1,774 @@
...
+    :returns: Workflow history
+    :rtype: obj
+    """
+    obj = get_object(brain_or_object)
+    review_history = []
+    try:
+        workflow = get_tool("portal_workflow")
+        review_history = workflow.getInfoFor(obj, 'review_history')
+    except WorkflowException:
+        return []
```
rockfruit: Probably a permission error, and so perhaps we should put it in the log for debugging

bika/lims/api.py
```diff
@@ -0,0 +1,774 @@
...
+    if rev is True:
+        review_history = reversed(review_history)
+    if not isinstance(review_history, (list, tuple)):
+        return list(review_history)
```
rockfruit: I don't see how review_history can be anything but an iterable; however if it is, I think we should wrap this line in try:except and fail if it can't be list()ed

bika/lims/api.py
```diff
@@ -0,0 +1,774 @@
...
+    """Get the version of the current object
+
+    :param brain_or_object: A single catalog brain or content object
+    :type brain_or_object: ATContentType/DexterityContentType/CatalogBrain
+    :returns: The current version of the object
+    :rtype: int
+    """
+    obj = get_object(brain_or_object)
+    if not is_versionable(obj):
+        return None
```
rockfruit: The caller should not be retrieving the version for an object that does not support it.  I believe this should fail too.

bika/lims/api.py
```diff
@@ -0,0 +1,774 @@
...
+def get_group(group_or_groupname):
+    """Return Plone Group
+
+    :param group_or_groupname: Plone group or the name of the group
+    :type groupname:  GroupData/str
+    :returns: Plone GroupData
+    :rtype: object
+    """
+    if not group_or_groupname:
+        return None
```
rockfruit: fail()

bika/lims/api.py
```diff
@@ -0,0 +1,774 @@
...
+def get_user(user_or_username):
+    """Return Plone User
+
+    :param user_or_username: Plone user or user id
+    :type groupname:  PloneUser/MemberData/str
+    :returns: Plone MemberData
+    :rtype: object
+    """
+    if not user_or_username:
+        return None
```
rockfruit: fail() :>

bika/lims/docs/API.rst
```diff
@@ -706,18 +706,20 @@ An empty property dict is returned if no user could be found::
...
+    >>> from operator import methodcaller
+
 Roles in Bika LIMS are basically a name for one or more permissions. For
 example, a `LabManager` describes a role which is granted the most permissions.
 
 To see which users are granted a certain role, you can use this function::
 
     >>> labmanagers = api.get_users_by_roles(["LabManager"])
-    >>> sorted(labmanagers)
+    >>> sorted(labmanagers, key=methodcaller('getId'))
```
ramonski: Cool 👍 I wasn't even aware that this function exists!

bika/lims/api.py
```diff
@@ -373,7 +367,7 @@ def get_object_by_path(path):
 
     res = pc(path=dict(query=path, depth=0))
     if not res:
-        return None
+        fail("Object at path '{}' not found".format(path))
```
ramonski: Why should we fail e.g. for `get_object_by_path("/Plone/foo")` and not just return None?
Which would just mean “Hey, there’s not such an object, go on"
That is the reason I do mostly `portal.get("foo", None)` instead of `portal["foo"]`
So the coder can decide to `assert(get_object_by_path("/Plone/foo"))` if it is really expected to have an object there, w/o having to do a `try..except`.

bika/lims/api.py
```diff
@@ -311,33 +308,30 @@ def get_object_by_uid(uid):
...
-    rc = get_tool("reference_catalog", None)
+    uc = get_tool("uid_catalog")
 
     # try to find the object with the reference catalog first
-    obj = rc and rc.lookupObject(uid)
-    if obj:
-        return obj
+    brains = uc(UID=uid)
+    if brains:
+        return brains[0].getObject()
```
ramonski: Cool, it definitely makes more sense to use the `uid_catalog` here.
But we should be consistent and do a `get_object(brains[0])` and fail if `len(brain) > 1`.

bika/lims/api.py
ramonski: Or maybe even better, we change the function signature to `def get_object_by_path(path, default=_marker)` and raise if no default value is given?

bika/lims/api.py
rockfruit: The UID catalog won't have >1 brains for a single UID!  If by some odd chance this occurs, they will certainly refer to the same object.

bika/lims/api.py
rockfruit: Then perhaps we should provide default=_marker as a parameter to this function, to allow caller to decide.

xispa:
I haven't got the opportunity to comment this PR before. Clean, elegant, beautiful code. For sure a key contribution, a model to follow. Thanks @ramonski

ramonski:
Thanks for your feedback @xispa, I'm happy that you find this code useful and when it helps us to write less code and have more time to enjoy life😄

PR #1949, opened 2017-02-16
Lims 2251 CLONE - AR Create. Sample Partitions' Container and Preservation not saved

AlcyonSuisse:
Fix issue with partitions, conservations and preservations not saved when creating AR with an ARTemplate or distinct container in AS.

https://jira.bikalabs.com/browse/LIMS-2251

--


bika/lims/utils/samplepartition.py
```diff
@@ -51,7 +51,7 @@ def create_samplepartition(context, data, analyses=[]):
     # Determine if the sampling workflow is enabled
     workflow_enabled = context.bika_setup.getSamplingWorkflowEnabled()
     # Sort containers and select smallest
-    container = data.get('container_uid', None)
+    container = data.get('container', None)
```
ramonski: I could not see where `data['container']` comes from. But if it is there, we should also consider to update the docstring of this function.

ramonski:
Ok, this is something we should have a look once again @rockfruit and try to merge your changes into `master`

PR #1951, opened 2017-02-21
Some small API changes

rockfruit:

Some small changes to bika.lims.api.  Commit log holds details.

--

### DIFF
--- a/bika/lims/api.py
+++ b/bika/lims/api.py
@@ -454,13 +454,15 @@ def get_parent(brain_or_object, catalog_search=False):
     return get_object(brain_or_object).aq_parent
 
 
-def search(query, catalog=_marker):
+def search(query, catalog=_marker, show_inactive=False):
     """Search for objects.
 
     :param query: A suitable search query.
     :type query: dict
     :param catalog: A single catalog id or a list of catalog ids
     :type catalog: str/list
+    :param show_inactive: Include inactive or dormant objects
+    :type show_inactive: Boolean
     :returns: Search results
     :rtype: List of ZCatalog brains
     """
@@ -507,6 +509,10 @@ def search(query, catalog=_marker):
 
     # With a single catalog, we don't have to care about merging the results
     if len(catalogs) == 1:
+        brains = catalogs[0](query)
+        # Avoid inactive or dormant items
+        if not show_inactive:
+            return filter(is_active, brains)
         return catalogs[0](query)
 
     # Multiple catalog results need to be merged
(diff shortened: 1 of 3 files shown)


bika/lims/api.py
```diff
@@ -497,7 +509,11 @@ def search(query, catalog=_marker):
 
     # With a single catalog, we don't have to care about merging the results
     if len(catalogs) == 1:
-        return catalogs[0](query)
+        brains = catalogs[0](query)
+        # Avoid inactive or dormant items
+        if not show_inactive:
+            return filter(is_active, brains)
+        return brains
```
ramonski: Would it make sense to just update the query instead of filtering the brains?

bika/lims/api.py
```diff
@@ -630,6 +657,28 @@ def get_workflow_status_of(brain_or_object):
     return ploneapi.content.get_state(obj)
 
 
+def is_active(brain_or_object):
+    """ Check if obj is inactive or cancelled.
+    """
```
ramonski: We should stick to the same docstring style as with the other functions

bika/lims/docs/API.rst
```diff
@@ -497,6 +514,21 @@ We could also explicitly define a catalog to achieve the same::
...
+    >>> results = api.search({"portal_type": "AnalysisCategory", "id": "analysiscategory-1"})
+    >>> len(results)
+    1
+    >>> transition(analysiscategory1, 'deactivate')
+    >>> results = api.search({"portal_type": "AnalysisCategory", "id": "analysiscategory-1"})
+    >>> len(results)
+    0
+    >>> results = api.search({"portal_type": "AnalysisCategory", "id": "analysiscategory-1"}, show_inactive=True)
+    >>> len(results)
+    1
```
ramonski: beautiful:) Maybe we could even proxy `transition` to our API. This would allow us to do some extra magic, like firing an Bika event or smth. like that

bika/lims/utils/__init__.py
```diff
@@ -116,18 +112,10 @@ def getUsers(context, roles, allow_empty=True):
...
-    wf = getToolByName(obj, 'portal_workflow')
-    if (hasattr(obj, 'inactive_state') and obj.inactive_state == 'inactive') or \
-       wf.getInfoFor(obj, 'inactive_state', 'active') == 'inactive':
-        return False
-    if (hasattr(obj, 'cancellation_state') and obj.inactive_state == 'cancelled') or \
-       wf.getInfoFor(obj, 'cancellation_state', 'active') == 'cancelled':
-        return False
-    return True
-
+    return api.is_active(object_or_brain)
```
ramonski: Nice! Thanks for adding that functionality into the API!
Maybe we could even add a depreciation warning in here?

PR #1958, opened 2017-03-01
Lims 2392 list and expire

Lunga001:
      Adding cron in plone for ReferenceSample.

https://jira.bikalabs.com/browse/LIMS-2392

## Current behavior before PR
       On the http://localhost:8080/Plone/@@cron-settings/add
       Name of the task : expireReferenceSamples
       UserID the task will run as: campbell
       How to repeat the job (CRON format): * * * * * * 
       
--


ramonski:
@Lunga001 `collective.cron` received the last update on 2013-02-18 and is currently in a non-usable state (at least for me).
I don't want to have this dependency in the `bika.lims` core and I would suggest to use the built in clock server of Zope (https://docs.plone.org/develop/plone/misc/asyncronoustasks.html) or better just provide an external method and let the users of Bika LIMS decide if they want to call it via a server cron job.

Lunga001:
@ramonski I've put on an external method (browser view)

bika/lims/adapters/configure.zcml
```diff
@@ -97,6 +97,7 @@
     <adapter name='bika.lims.IHaveIdentifiersModifier' factory=".identifiers.IHaveIdentifiersSchemaModifier"/>
 
     <adapter factory=".identifiers.IdentifiersIndexer" name="Identifiers"/>
+    <adapter factory=".cron.ExpireReferenceSamples" name="expireReferenceSamples"/>
```
ramonski: Please remove

bika/lims/browser/expirereferencesamples.py
```diff
@@ -0,0 +1,44 @@
...
+from plone import api
+
+from bika.lims import logger
+from bika.lims.browser import BrowserView
+from bika.lims import bikaMessageFactory as _
+
+class ExpireReferenceSamplesView(BrowserView):
+    """Expireference Samples
+    """
+
```
ramonski: Maybe just add a `__call__` method in here, so we can call it like `http://127.0.0.1:8080/bikalims/@@ expire_reference_samples `

bika/lims/browser/expirereferencesamples.zcml
```diff
@@ -0,0 +1,16 @@
...
+    i18n_domain="bika">
+
+    <browser:page
+      for="*"
+      name="expire_reference_samples"
+      class="bika.lims.browser.expirereferencesamples.ExpireReferenceSamplesView"
+      permission="cmf.ManagePortal"
+      attribute="expire_reference_samples"
+      layer="bika.lims.interfaces.IBikaLIMS"
+    />
```
ramonski: Ok, this looks like a feasible way.
I think an external Cronjob could now do something like this:

```
# m h  dom mon dow   command

*/5 * * * * wget --delete-after http://admin:password@127.0.0.1:8081/bikalims/@@expire_reference_samples
```

bika/lims/profiles/default/metadata.xml
```diff
@@ -2,6 +2,7 @@
 <metadata>
   <version>320</version>
   <dependencies>
+    <dependency>profile-collective.cron:default</dependency>
```
ramonski: Please remove

bika/lims/browser/expirereferencesamples.py
```diff
@@ -0,0 +1,44 @@
...
+        query = {'portal_type': 'ReferenceSample',
+                 'getExpiryDate': {'query': datetime.today(),
+                                   'range': 'max'},
+                 'review_state': 'current',
+                }
+        brains = bc(query)
+        cnt = 0
+        for brain in brains:
+            obj = brain.getObject()
+            state = api.content.get_state(obj=brain.getObject())
```
ramonski: Why do you need the `state` here? Also you have already the object, so there is no need to do `brain.getObject()` again.

bika/lims/browser/expirereferencesamples.py
```diff
@@ -0,0 +1,44 @@
...
+            try:
+                api.content.transition(obj=obj, transition='expire')
+                cnt += 1
+            except Exception, e:
```
ramonski: Please look up the exception that a workflow transition raises (something like WorkflowException) and catch that explicitly. Just for better coding style.

bika/lims/browser/referencesample.py
```diff
@@ -410,10 +422,11 @@ def folderitem(self, obj, item, index):
                     item['review_state'] = 'expired'
                     item['obj'] = obj
 
-        if self.contentFilter.get('review_state', '') \
-           and item.get('review_state', '') == 'expired':
-            # This item must be omitted from the list
-            return None
+        if self.review_state['id'] != 'expired':
+            if self.contentFilter.get('review_state', '') \
+               and item.get('review_state', '') == 'expired':
```
ramonski: Decrease complexity of this function (level of indentation), e.g.:

```python
review_state = item.get("review_state")
content_filter = self.contentFilter.get('review_state', False)

if self.review_state['id'] == 'expired':
    return None
if review_state == "expired" and content_filter:
    return None
```

bika/lims/browser/expirereferencesamples.py
Lunga001: Found this exception InvalidParameterError on a transition from the plone api

bika/lims/browser/expirereferencesamples.zcml
Lunga001: Yes

bika/lims/browser/referencesample.py
Lunga001: This brings a different behaviour and so I've done 
       if self.review_state['id'] != 'expired' \                               
               and review_state == "expired" and content_filter:               
           return None

bika/lims/browser/expirereferencesamples.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+
+    def __call__(self):
+        bc = api.portal.get_tool('bika_catalog')
+        query = {'portal_type': 'ReferenceSample',
+                 'getExpiryDate': {'query': datetime.today(),
+                                   'range': 'max'},
+                 'review_state': 'current',
+                }
+        brains = bc(query)
```
rockfruit: This query doesn't seem to find the correct reference samples for me.  It seems the date query is failing.  I think you should be using DateTime.DateTime from zope

bika/lims/browser/expirereferencesamples.zcml
rockfruit: @Lunga001  I think include the above text in documentation; add bika/lims/docs/cron-scripts.rst - we can collect docs on other cron views there later, too.

Thanks

ramonski:
@Lunga001: Can you please re-visit this PR an give add a comment how we should continue with it? Thanks.

PR #1959, opened 2017-03-02
LIMS-2589: Commented out User activity report

Lunga001:
      Remove user activity report
https://jira.bikalabs.com/browse/LIMS-2589

## Current behavior before PR
     It was taking too long to run the activity report
     Taken out the user activity report link
--

### DIFF
--- a/bika/lims/browser/reports/templates/administration.pt
+++ b/bika/lims/browser/reports/templates/administration.pt
@@ -129,6 +129,7 @@
                 </li>
             </ul>
 
+            <!--
             <h2 i18n:translate="">Traceability</h2>
             <ul id="toc-reports">
                 <li>
@@ -160,6 +161,7 @@
                     </div>
                 </li>
             </ul>
+            -->
         </fieldset>
     </div>
 </metal:content-core>


ramonski:
@xispa, @rockfruit: do we want to generally disable this feature on `master`? Not sure if this is only a customer requirement or not.

ramonski:
After some thinking I see this change more in a customer Add-on than being merged into `master`, where it affects all of us.

I also disagree to eliminate existing Bika LIMS functionality, because of performance issues.
We should rather analyze where the performance issue comes from and then improve it, as many other clients may use this activity report or depend on it.

Therefore, I reject this PR and close it now. Thanks for understanding.

lemoene:
Understood. In this case, the report kills the system, and causes support queries. I don't know of anybody using it as it is practically 'ueseless'. Until it is working properly I requested that it be commented out of UI - (unless we use JSON:) I understood this will be complicated to fix and we do not have capacity at the mo.

Somebody at ALS is going to push that button sooner or later, I bet Paul.

ramonski:
Thanks for your comment Lemoene. Just eliminate this report in the ALS specific package, maybe just with something as simple as a CSS `display: none` rule.

Removing existing features should be generally handled with care, as we don't know who of our clients is using it and might get angry if we do so.

PR #1960, opened 2017-03-02
Lims 2488 ar import spreadsheet improvements

mikejmets:

https://jira.bikalabs.com/browse/LIMS-2488

AR Import Spreadsheet now take a Sampler field and the DateSampled is now a DateTime field. 

--


bika/lims/content/arimport.py
```diff
@@ -352,6 +356,20 @@ def workflow_script_import(self):
...
+                found = False
+                users = getUsers(self, ['LabManager', 'Sampler']).items()
+                for (samplerid, samplername) in users:
+                    if row['Sampler'] == samplername:
+                        found = True
+                        row['Sampler'] = samplerid
+                        break
+                if not found:
+                    #TODO Not sure what to do
+                    raise RuntimeError(
```
mikejmets: Please advise on how to fail 'nicely' here

bika/lims/tests/test_ARImport.py
```diff
@@ -140,8 +140,8 @@ def test_complete_valid_batch_import(self):
             self.fail('12 Analysis not found! We found %s' % l)
         states = [workflow.getInfoFor(a.getObject(), 'review_state')
                   for a in analyses]
-        if states != ['sample_due'] * 12:
```
mikejmets: I'm concerned that we changed the test to reflect the result and I don't know why.

bika/lims/tests/test_ARImport.py
```diff
@@ -217,6 +217,85 @@ def test_LIMS_2081_post_edit_fails_validation_gracefully(self):
...
+        l = len(analyses)
+        if l != 12:
+            self.fail('12 Analysis not found! We found %s' % l)
+        states = [workflow.getInfoFor(a, 'review_state')
+                  for a in analyses]
+        if states != ['sampled'] * 12:
+            self.fail('Analysis states should all be sampled, but are not!')
+        for a in analyses:
+            workflow.doActionFor(a, 'sample_due')
+        states = [workflow.getInfoFor(a, 'review_state')
```
mikejmets: This new test goes one step further than the "valid" test. Should this be retro fitted into it?

bika/lims/content/arimport.py
```diff
@@ -352,6 +356,20 @@ def workflow_script_import(self):
...
             row['Contact'] = self.getContact()
+            row['DateSampled'] = row['DateSampled'].replace('-', '/')
+            if row['Sampler']:
+                #Lookup sampler's uid
+                found = False
+                users = getUsers(self, ['LabManager', 'Sampler']).items()
+                for (samplerid, samplername) in users:
+                    if row['Sampler'] == samplername:
+                        found = True
+                        row['Sampler'] = samplerid
```
ramonski: Please keep in mind, that the fullname is not unique amongst the users. Therefore, I'm not sure if this whole logic makes sense, that we just take the first user found.

A better approach would be to search first the user by ID, then by Fullname, bail out if we get multiple results and then maybe check the roles of this user. Please also consider to refactor out this functionality into an own function/method for better readability.

bika/lims/content/arimport.py
```diff
@@ -352,6 +356,20 @@ def workflow_script_import(self):
...
+                        break
+                if not found:
+                    #TODO Not sure what to do
+                    raise RuntimeError(
+                                'Import sampler %s unknown' % row['Sampler'])
```
ramonski: If the import depends on it, I would go for a `ValueError` if the `findUser` function returned `None`. But if I remember correctly, the user has afterwards the possibility to manually set this field with a value. So maybe we should postpone the Error raising when the "real" data import is done.

bika/lims/content/arimport.py
```diff
@@ -352,6 +356,20 @@ def workflow_script_import(self):
             row['ClientReference'] = self.getClientReference()
             row['ClientOrderNumber'] = self.getClientOrderNumber()
             row['Contact'] = self.getContact()
+            row['DateSampled'] = row['DateSampled'].replace('-', '/')
```
ramonski: Please consider to call a function, e.g. `convert_date_string(row['DateSampled'])`. This makes it easier to test later.

bika/lims/tests/test_ARImport.py
```diff
@@ -140,8 +140,8 @@ def test_complete_valid_batch_import(self):
             self.fail('12 Analysis not found! We found %s' % l)
         states = [workflow.getInfoFor(a.getObject(), 'review_state')
                   for a in analyses]
-        if states != ['sample_due'] * 12:
-            self.fail('Analysis states should all be sample_due, but are not!')
+        if states != ['sampled'] * 12:
```
ramonski: Hmm, I'm not sure _why_ the states changed now from `sample_due` to `sampled`, just because we added the Sample field. @rockfruit can you explain?

bika/lims/tests/test_ARImport.py
```diff
@@ -217,6 +217,85 @@ def test_LIMS_2081_post_edit_fails_validation_gracefully(self):
...
+            errors = arimport.getErrors()
+            self.fail(
+                'Importation failed!  %s.Errors: %s' % (arimport.id, errors))
+
+        bc = getToolByName(self.portal, 'bika_catalog')
+        ars = bc(portal_type='AnalysisRequest')
+        if not ars[0].getObject().getContact():
+            self.fail('No Contact imported into ar.Contact field.')
+        l = len(ars)
+        if l != 4:
```
ramonski: Merge the two lines to `if len(ars) is not 4:`

bika/lims/tests/test_ARImport.py
ramonski: I'm not sure what you mean with that. Could you explain please?

bika/lims/content/arimport.py
mikejmets: Done

bika/lims/tests/test_ARImport.py
mikejmets: I'm not sure either. Please ignore

bika/lims/content/arimport.py
mikejmets: Done

bika/lims/content/arimport.py
mikejmets: Agreed. I created a method that raises ValueError. Then commented them out and replaced them with "return None". So if anyone feels strongly about raising the errors, the code is there and can be uncommented.

bika/lims/tests/test_ARImport.py
mikejmets: The variable is set because it is used in the fail message in the line below.

Issue #1992, opened 2017-03-22
Data Duplication between Analysis Requests and Samples


ramonski:
## Description

The current architecture of `bika.lims` couples AnalysisRequests tightly to Samples, so that e.g. a sample will be created automatically when an AR is created and parts of the information is managed through the managing AR.

Therefore, some of the fields are as well on `AnalysisRequest` and `Sample` present and custom `setter` methods on the AR set their field values transparently on the underlying (managed) Sample.

The issue here is now, that the data is set on the AR as well as on the Sample and _only_ if the Data was set through the AR. Setting such a field on the Sample object will lead to inconsistent data, because the managing AR never gets notified by this change.

## Proposal

We need a transparent "Proxy Field", which sets and gets the data from the underlying sample and do not store the value by it's own. Thus, the data can be changed by the AR or the Sample and still keeps consistent.

PR #1993, opened 2017-03-23
Implemented Proxy Field to fix data duplication between ARs and Samples

ramonski:

Please see https://github.com/bikalabs/bika.lims/issues/1992 for details.

## Current behavior before PR

- Data is duplicated on AR and Sample
- Setting some fields on Sample conducted to data inconsistency between AR and Sample

- Data is not duplicated between AR and Sample
- Setting the data on AR or Sample keeps the data always consistent

--

### DIFF
--- a/bika/lims/api.py
+++ b/bika/lims/api.py
@@ -9,6 +9,7 @@ from Products.CMFPlone.utils import base_hasattr
 from Products.CMFCore.interfaces import ISiteRoot
 from Products.Archetypes.BaseObject import BaseObject
 from Products.ZCatalog.interfaces import ICatalogBrain
+from Products.CMFPlone.utils import _createObjectByType
 from Products.CMFCore.WorkflowCore import WorkflowException
 
 from zope import globalrequest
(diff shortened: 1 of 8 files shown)


ramonski:
The current failure in the test happens because the Sample is not set at the time where the value for the proxy fields, e.g. 'SampleType', are set.

=> This assumes that creating ARs from the current JSON API only set the values **only** on the AR but **not** on the Sample.

ramonski:
This PR one seems to me really important, because the current data duplication (AR contains same fields as Sample) covers an error, when the AR does not have a Sample set during setting these fields.

While this seems only be possible in code, e.g. through the current JSON API create, it should be generally avoided, because it creates inconsistent data in the DB and that is bad for a LIMS.

ramonski:
Behavior after this PR is merged:

- ARs contains Proxy fields, which evaluate the field expression (`context.getSample()`) to get the proxied object.
- If the proxied object could not be found (e.g. `context.getSample()` returns `None`), an error message is logged and neither the AR proxy field (which cannot hold own data) nor any other field is set => Field returns `None`
- If the proxied object is found, but does **not** contain a field similar named to the proxy field, an error is raised

N.B. To return `None` if the proxied object could not be found is necessary to maintain backwards compatibility to the JSONAPI create functionality.

ramonski:
Ok, this one is ready to be merged. Please see included doctest for more details about the functionality: https://github.com/bikalabs/bika.lims/pull/1993/files#diff-882a02198e8802742b8bdc7a192ef4bb

PR #2003, opened 2017-03-29
Lims 2371 rounding

zylinx:
https://jira.bikalabs.com/browse/LIMS-2371

## Current behavior before PR
Only decimal / scientific notation, round is only done to decimal precision.

Rounding can be done on significant figures.

We need to sort out the reporting of results to three significant figures as well as rounding of results. 
One of the issues we had with our previous LIMS is that the system would not report to three significant figures.

So if we had an instrument connected, interfaced with LIMS, and the raw result was 101.123456789mg/L, 
this is the result that appeared on the report (which is wrong  - the result should be only 101mg/L – three 
significant figures. I’ve been adding some results to the BIKA (see below) – I can see the system correcting 
to three significant figures?

Is this something we can sort out in the BIKA setup or this feature is currently not available (I’ll have to get t
he staff to manually change every result that is not reported correctly to three significant figures) – reporting 
to three significant figures  is an ISO17025 requirement.

--


ramonski:
How shall we continue with this one @rockfruit?

Issue #1973, opened 2017-03-14
Published PDF reports are visible by anonymous users


ramonski:
## Steps to reproduce

1. Publish an AR for `client-1`
2. Log out of the System
3. Navigate to this URL: `<bika-host>/clients/client-1/<AR ID>/arreport-1/at_download/Pdf`

## Current behavior

The PDF is viewable by anonymous users

## Expected behavior

Permission should be denied

ramonski:
We need to set a different Workflow for `ARReport` as they have the `plone_workflow` assigned and have per default a Status of `Public draft` after publication.

PR #2043, opened 2017-05-17
Issue-1973: Published AR reports are viewable by anonymous

ramonski:

Please see https://github.com/bikalabs/bika.lims/issues/1973 for details

--


bika/lims/profiles/default/workflows.xml
```diff
@@ -91,6 +91,11 @@
       <bound-workflow workflow_id="bika_cancellation_workflow"/>
     </type>
 
+    <!-- AR Reports -->
```
rockfruit: Woops.  We should apply a default workflow with much stricter permissions.  This is how private data ends up on google!

PR #2058, opened 2017-05-23
Unpinned openpyxl

mikejmets:
Campbell says "openpyxl recently added a workaround code for systems where lxml version is < 3.0 (like plone 4).  that was the reason that this code was in a PR and never merged" 

## Current behavior before PR
openpyxl was pinned to 1.5.8 which caused problems with the export and import scripts

Export and import scripts function correctly.

--


mikejmets:
This is causes a problem when loading setup data on a clean site.

mikejmets:
New version of OpenPyXL behaves differently when opening a sheet that doesn't exist so I added a check.

mikejmets:
Relevant changes are in PR-2071

PR #2110, opened 2017-06-22
Bc 69 client contact department lookup

Lunga001:

https://jira.bikalabs.com/browse/BC-69
https://jira.bikalabs.com/browse/BC-68

A Client can have license from government agencies to operator in a certain capacity (grow, sell, distribute a commodity) so this is an extension to client data to keep those license number in the database.

## Current behavior before PR
      On the client contact the Department field is not a lookup, it's text.
     Clients Departments on control panel to allow adding a Client Department
     The Client Department will be used to lookup the Client Contact Department field
     Clients Types on control panel to allow adding a Client Type
     The Client Type will be used to lookup the Licenses on a Client's Licenses field
--


mikejmets:
Lunga is busy adding a doctest

Lunga001:
Doctest added

mikejmets:
I'm concerned about client departments not being indexed in bika_setup_catalog.

bika/lims/content/client.py
```diff
@@ -130,6 +136,33 @@
...
+        schemata = "Licenses",
+        allow_insert=True,
+        allow_delete=True,
+        allow_reorder=True,
+        allow_empty_rows=False,
+        columns=('LicenseType',
+                 'MME_ID',
+                 'MME_Registration',
+                 'Authority'),
+        default=[{'LicenseType': '',
```
ramonski: What does these values mean? We're shipping a system w/o documentation, so the system needs to speak to the user (and the other devs). Please add a description to the widget, so that it tells the user in the UI what this means.

bika/lims/content/clientdepartment.py
```diff
@@ -0,0 +1,86 @@
...
+
+class IClientDepartment(model.Schema):
+    """ A Client Departments container.
+    """
+
+class ClientDepartment(Item):
+    implements(IClientDepartment)
+    pass
+
+    # Catalog Multiplex support
```
ramonski: All methods below feel like a hack. Why can't we do this in an event handler for object added/modified?

bika/lims/content/clientdepartment.py
```diff
@@ -0,0 +1,86 @@
...
+                c.reindexObject(self,
+                                idxs=idxs)
+
+@indexer(IClientDepartment, IBikaSetupCatalog)
+def clienttype_title_indexer(obj):
+    if obj.title:
+        return obj.title
+
+@indexer(IClientDepartment, IBikaSetupCatalog)
+def clienttype_sortable_title_indexer(obj):
```
ramonski: not sure why this index is needed. Does it make a difference if the title only is indexed including whitespaces or without?

bika/lims/content/clienttype.py
```diff
@@ -0,0 +1,76 @@
...
+from plone.indexer import indexer
+from plone.supermodel import model
+from Products.CMFCore.utils import getToolByName
+from zope.interface import implements
+from zope import schema
+
+class ClientType(Item):
+    implements(IClientType)
+
+    # Catalog Multiplex support
```
ramonski: same here. Please keep also the DRY (don't repeat yourself) rule in mind

bika/lims/content/contact.py
```diff
@@ -60,11 +64,30 @@
...
+    StringField(
+        'Department',
+        mode="rw",
+        read_permission=permissions.View,
+        write_permission=permissions.ModifyPortalContent,
+        vocabulary='getClientDepartment',
+        acquire=True,
+        widget=SelectionWidget(
+            format="select",
+            label=_("Department"),
```
ramonski: please add a speaking description

bika/lims/controlpanel/bika_clientdepartments.py
```diff
@@ -0,0 +1,89 @@
...
+             'title': _('All'),
+             'contentFilter':{},
+             'columns': ['Title',
+                         'Description',
+                         ]},
+        ]
+
+    def folderitems(self):
+        items = BikaListingView.folderitems(self)
+        for x in range(len(items)):
```
ramonski: Please don't use this pattern that accesses the values by index.
Either use `for item in items` or the `folderitem` method, which is called per item.

bika/lims/controlpanel/bika_clienttypes.py
```diff
@@ -0,0 +1,86 @@
...
+             'title': _('All'),
+             'contentFilter':{},
+             'columns': ['Title',
+                         'Description',
+                         ]},
+        ]
+
+    def folderitems(self):
+        items = BikaListingView.folderitems(self)
+        for x in range(len(items)):
```
ramonski: same here

bika/lims/exportimport/setupdata/__init__.py
```diff
@@ -2307,3 +2308,23 @@ def Import(self):
...
+                        title=row['title'], 
+                        description=row.get('description', ''))
+
+class Client_Types(WorksheetImporter):
+
+    def Import(self):
+        folder = self.context.bika_setup.bika_clienttypes
+        for row in self.get_rows(3):
+            if row['title']:
+                obj = api.content.create(folder, 'ClientType',
```
ramonski: we should better use `bika.lims.api.create` for creating contents, as there are some more things to take care when creating contents in Bika LIMS.

bika/lims/upgrade/to340.py
```diff
@@ -19,15 +19,23 @@
...
-    """Upgrade step to prepare for refactored ID Server
+    """Upgrade step to 3.4.0
     """
     portal = aq_parent(aq_inner(tool))
 
     qi = portal.portal_quickinstaller
     ufrom = qi.upgradeInfo('bika.lims')['installedVersion']
     logger.info("Upgrading Bika LIMS: %s -> %s" % (ufrom, '3.4.0'))
 
+    pc = getToolByName(portal, 'portal_catalog')
```
ramonski: Seems like the portal_catalog is not needed here.

bika/lims/content/client.py
mikejmets: Renamed the Fields to be non BC specific and added a description

bika/lims/content/clientdepartment.py
mikejmets: I don't think is can be done in an event handler be cause we are overidding the methods in the superclass Item. I know this is a hack. It comes out a oneliner from Lemoene "all new content types should be dexterity". That's easy enough until you have multiple catalogs. Maybe it can be done with an adapter but I don't know how to do it.

bika/lims/content/clientdepartment.py
mikejmets: sortable_title is index by default in Archetypes and Dexterity so it must be done in this multiple catalog override.

bika/lims/content/clienttype.py
mikejmets: Sure, but don't know how to do it.

bika/lims/controlpanel/bika_clientdepartments.py
mikejmets: This is a cut/paste job from the other controlpanel classes. Can you point me to one that you have  refactored correctly?

bika/lims/controlpanel/bika_clienttypes.py
mikejmets: Ditto above

bika/lims/exportimport/setupdata/__init__.py
mikejmets: Does it handle dexterity?

bika/lims/upgrade/to340.py
mikejmets: True

bika/lims/controlpanel/bika_clientdepartments.py
ramonski: You could do it either like this:
https://github.com/bikalabs/bika.lims/blob/master/bika/lims/controlpanel/bika_instrumentlocations.py#L74

Or implement the `folderitem` method:
https://github.com/bikalabs/bika.lims/blob/master/bika/lims/browser/bika_listing.py#L814

Which is called by the original `folderitems` method here:
https://github.com/bikalabs/bika.lims/blob/master/bika/lims/browser/bika_listing.py#L1049

Anyhows, I really don't like the way how this bika listing view works, because it always includes object information, which are expensive to fetch from the ZODB, just in case that the user might want it (eventually) in the future.
It would be much more faster and better to let the user decide when he/she would like to get the full object information with a separate JSON request instead of maintaining this giant dictionary object... Sorry, off-topic, but I always get emotional with that listing view and always when I see it, I want to change this code...

bika/lims/exportimport/setupdata/__init__.py
ramonski: I belive yes, but I might overlook something. Please try it out and give feedback.

bika/lims/content/clienttype.py
ramonski: Please check out this documentation:
https://docs.plone.org/4/en/external/plone.app.dexterity/docs/advanced/event-handlers.html

Basically you need event handlers for `zope.lifecycleevent.interfaces.IObjectCreatedEvent` and `zope.lifecycleevent.interfaces.IObjectModifiedEvent` and register them via ZCML:

```
<subscriber
    for="bika.lims.interfaces.IClientType zope.lifecycleevent.interfaces.IObjectCreatedEvent"
    handler="bika.lims.subscribers.catalogObject"
    />
```

bika/lims/content/clientdepartment.py
ramonski: ok, I was just wondering about the `[w.lower() for w in obj.title.split(' ')]` dance that have to be performed in this index...

bika/lims/controlpanel/bika_clientdepartments.py
mikejmets: Sorry, I missed you previous point. I much prefer your approach.

bika/lims/controlpanel/bika_clienttypes.py
mikejmets: Done

bika/lims/content/contact.py
mikejmets: Done

bika/lims/exportimport/setupdata/__init__.py
mikejmets: Tried it, it doesn't work. It uses _createObjectByType. It should maybe use plone api.

bika/lims/content/clienttype.py
mikejmets: Done. Thanks for insisting on this.

bika/lims/exportimport/setupdata/__init__.py
ramonski: Thanks, good to know. Need to be properly changed in the `bika.lims.api`

bika/lims/subscribers/catalogobject.py
```diff
@@ -0,0 +1,86 @@
...
+from bika.lims.permissions import ManageSupplyOrders, ManageLoginDetails
+
+
+def indexObject(obj, event):
+    """ Various types need automation on edit.
+    """
+    if not hasattr(obj, 'portal_type'):
+        return
+
+    if obj.portal_type not in ('ClientType', 'ClientDepartment'):
```
ramonski: I would like to see a more general approach in here, instead of having the contents listed here explicitly.
You could for example add 3 more attributes to `ClientType` and `ClientDepartment` that contain a list of catalog names, e.g. `index_catalogs = ['bika_setup_catalog', 'portal_catalog']`. Then simply do `map(lambda c: api.get_tool(c).catalog_object(obj), obj.index_catalogs)` in the event handler and it can be simply used by any other Dexterity object as well.

PR #2124, opened 2017-07-09
BC-83 Select Template dropdown now shows Worksheet templates

Lunga001:

https://jira.bikalabs.com/browse/BC-83
https://github.com/bikalabs/bika.lims/issues/

## Current behavior before PR
     Worksheet Templates not listed for WS Create
     On the Select template drop down show the Worksheet Templates
--

### DIFF
--- a/bika/lims/browser/worksheet/views/folder.py
+++ b/bika/lims/browser/worksheet/views/folder.py
@@ -459,7 +459,11 @@ class FolderView(BikaListingView):
         """ List of templates
             Used in bika_listing.pt
         """
-        return DisplayList(self.templates)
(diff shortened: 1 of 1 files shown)


Lunga001:
The query that's done on the __init__ function(browser/worksheet/views/folder.py) that gets self.templates returns an empty list but the same query that's done getWorksheetTemplates function returns something. It seems like queries cannot be made inside the __init__ function

ramonski:
Thanks @Lunga001 for the PR. Indeed, catalog queries shouldn't be done in the `__init__` method of a browser view, because they will be done as `Anonymous` user. Also the ZCA (adapter lookups etc.) shouldn't be done there.

Please don't forget to update the Changelog the next time. Well done.

PR #2187, opened 2017-08-19
AR Add 2: Prevent attempts to set missing fields

rockfruit:

If an extension package has hidden fields from the AR-Add form, this can result in set_reference_field being called on missing fields.

## Current behavior before PR

AR Add form spinner appears when modifying some reference field (eg, the Sample Type field will cause this to happen if the DefaultContainerType field is hidden by an AT Schema Extender).  A message claiming that "existing_uids is not defined" appears in console.log, and the form cannot be submitted.

This little defensive check causes the function to return early if required field element is missing.

--

### DIFF
--- a/bika/lims/browser/js/coffee/bika.lims.analysisrequest.add.coffee
+++ b/bika/lims/browser/js/coffee/bika.lims.analysisrequest.add.coffee
@@ -389,6 +389,11 @@ class window.AnalysisRequestAdd
 
     me = this
     $field = $(field)
+
+    # If the field doesn't exist in the form, avoid trying to set it's values
+    if ! $field.length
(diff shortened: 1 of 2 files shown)


bika/lims/browser/js/coffee/bika.lims.analysisrequest.add.coffee
```diff
@@ -389,6 +389,11 @@ class window.AnalysisRequestAdd
 
     me = this
     $field = $(field)
+
+    # If the field doesn't exist in the form, avoid trying to set it's values
+    if ! $field.length
```
ramonski: Cool, thanks!
In CS you could also writes something like this:
```
if $field.length isnt 1
```
or something like
```
return unless field.length
```

PR #2197, opened 2017-09-01
Fixed Bika listing cache key to include the review_state

ramonski:

Part of Issue https://github.com/bikalims/bika.lims/issues/2178

## Current behavior before PR

The Bika listing `cache_key` did only calculate the `UID` and the `modified` values. Unfortunately a WF transition leave the modification date unchanged.

From Slack:
<hr/>
mikemets [16:01] 
Recently we introduced some caching on the bika_listing and today it bit us in the butt. The cache key contained the method, uid and modified date. Unfortunately plone doesn't seem to update the modified date after transitioning on object's workflow. So we added the state to the key.  This works but looking up the state is not quick so I thought of adding an AfterTransition subscriber that does update modified. Anyone got any other ideas?
<hr/>

The brain is passed into the `cache_key` function to calculate the `review_state` as well.

--

### DIFF
--- a/bika/lims/api.py
+++ b/bika/lims/api.py
@@ -29,6 +29,7 @@ from zope.lifecycleevent import ObjectCreatedEvent
 from zope.security.interfaces import Unauthorized
 
 from plone import api as ploneapi
+from plone.memoize.volatile import DontCache
 from plone.api.exc import InvalidParameterError
 from plone.dexterity.interfaces import IDexterityContent
 from plone.app.layout.viewlets.content import ContentHistoryView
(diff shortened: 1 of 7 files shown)


ramonski:
The actual issue in that case is not the cache key itself, but the traceback raised when doing the "submit" transition in  https://github.com/bikalims/bika.lims/blob/master/bika/lims/workflow.py#L53.
This traceback is the reason, why the modification timestamp get rolled-back or never set.

ramonski:
Linked to Issue https://github.com/bikalims/bika.lims/issues/2198

ramonski:
A WF change of an AR or of any contained objects (Analyses, Attachments, Partitions etc.) will cause an update of the modification timestamps of the whole structure. This ensures that the Cache key (which uses the modification date) invalidates and the contents are re-fetched.

3 custom events are introduced in the API module. `IBikaBeforeTransitionEvent` will be triggered _before_ the WF transition, `IBikaAfterTransitionEvent` after the WF transition and `IBikaTransitionFailedEvent` if the transition failed (possible if multiple WFs are attached to an object). These events are called for all transitions that use either `bika.lims.workflow.doActionFor` or `bika.lims.api.do_transtion_for`.
All manually triggered transitions will obviously not notify these events.

Issue #2218, opened 2017-09-13
Cache Key does not invalidate if proxied Sample values of ARs change


ramonski:
## Steps to reproduce

1. Visit the AR listing view of a Client (make sure the column Client Reference is displayed)
2. Click on one of the listed ARs
3. Change the Client Reference Value of the AR
4. Navigate back to the AR listing view of the Client

## Current behavior

The AR listing view still shows the old value for the Client Reference

## Expected behavior

The AR listing view updates its values accordingly

## Screenshot (optional)

<img width="1206" alt="h2o17-0067-r1_ _bika" src="https://user-images.githubusercontent.com/713193/30390086-3fee3768-98b5-11e7-8dad-faad74fe5136.png">

<img width="1209" alt="riding_bytes_ _bika" src="https://user-images.githubusercontent.com/713193/30390231-ab4e9958-98b5-11e7-8936-f145f5d4cb5d.png">

ramonski:
Still an issue. We need to pass in the full object into https://github.com/bikalims/bika.lims/blob/master/bika/lims/browser/bika_listing.py#L874 to get the **real**  modification date w/o the need to reindex the catalog brain.

ramonski:
And finally ... it works;)

PR #2221, opened 2017-09-13
Issue-2218: Update modification dates when setting a proxy field

ramonski:

Please see https://github.com/bikalabs/bika.lims/issues/2218 for details.

## Current behavior before PR

Modification dates were not updated when a value was set on a proxy field

Modification date is set on proxy and holder object.

--

### DIFF
--- a/bika/lims/browser/fields/proxyfield.py
+++ b/bika/lims/browser/fields/proxyfield.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+import DateTime
+
 from zope.interface import implements
 
 from AccessControl import ClassSecurityInfo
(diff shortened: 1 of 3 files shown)


bika/lims/browser/fields/proxyfield.py
```diff
@@ -97,6 +99,15 @@ def set(self, instance, value, **kwargs):
         # set the value on the proxy object
         field.set(proxy_object, value, **kwargs)
 
+        # get the current time
+        now = DateTime.DateTime()
+
+        # update the modification date of the proxied object
+        proxy_object.setModificationDate(now)
```
ramonski: @rockfruit, @mikejmets: I don't know why the test below fails if I don't do this. Shouldn't the modification date be updated on the Sample because of the `field.set` in Line 100?

bika/lims/browser/fields/proxyfield.py
```diff
@@ -97,6 +99,15 @@ def set(self, instance, value, **kwargs):
...
+
+        # update the modification date of the holding object
+        instance.setModificationDate(now)
```
ramonski: Also I don't know why I can't do that instead:

```
from zope.lifecycleevent import modified

modified(instance)
```

PR #2228, opened 2017-09-14
LIMS-2695: Improve performance of worksheet results view

rockfruit:
(cherry picked from commit 7fa47a7e99bbcf1512319b37815cb18591f2dfc2)

Cache some expensive function calls, including historyawarereferencefield/get.

https://jira.bikalabs.com/browse/LIMS-2695

## Current behavior before PR

Worksheet manage_results view is very slow

performance is acceptable

--

### DIFF
--- a/bika/lims/browser/fields/historyawarereferencefield.py
+++ b/bika/lims/browser/fields/historyawarereferencefield.py
@@ -19,9 +19,20 @@ from Products.CMFEditions.Permissions import SaveNewVersion
 from Products.CMFEditions.Permissions import AccessPreviousVersions
 from Products.Archetypes.config import REFERENCE_CATALOG
 from bika.lims import bikaMessageFactory as _
+from bika.lims.api import bika_cache_key_decorator, get_uid
 from bika.lims.utils import t
 from bika.lims import logger
 from bika.lims.utils import to_utf8
+from plone.memoize.volatile import store_on_context, cache, DontCache
+
+
+def cache_key(method, field, instance, **kwargs):
+    creation_flag = instance.checkCreationFlag()
+    if creation_flag:
+        raise DontCache
+    uid = get_uid(instance)
+    versions = ''.join(instance.get('reference_versions', {}).values())
+    return "{}-{}".format(uid, versions)
 
 
 class HistoryAwareReferenceField(ReferenceField):
@@ -127,6 +138,7 @@ class HistoryAwareReferenceField(ReferenceField):
 
     security.declarePrivate('get')
 
+    @cache(cache_key)
     def get(self, instance, aslist=False, **kwargs):
         """get() returns the list of objects referenced under the relationship.
         """
--- a/bika/lims/content/instrument.py
+++ b/bika/lims/content/instrument.py
@@ -13,6 +13,8 @@ from Products.CMFCore.utils import getToolByName
 from Products.CMFPlone.utils import safe_unicode
 from Products.Archetypes.atapi import DisplayList, PicklistWidget
 from Products.Archetypes.atapi import registerType
+from bika.lims.api import bika_cache_key_decorator
+from plone.memoize.volatile import store_on_context, cache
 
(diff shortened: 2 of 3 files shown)


bika/lims/browser/fields/historyawarereferencefield.py
```diff
@@ -19,9 +19,20 @@
...
 from Products.Archetypes.config import REFERENCE_CATALOG
 from bika.lims import bikaMessageFactory as _
+from bika.lims.api import bika_cache_key_decorator, get_uid
 from bika.lims.utils import t
 from bika.lims import logger
 from bika.lims.utils import to_utf8
+from plone.memoize.volatile import store_on_context, cache, DontCache
+
+
+def cache_key(method, field, instance, **kwargs):
```
mikejmets: We learn some hard lessons (like on live) about must be included in the cache key. I would suggest you use api.get_cache_key and then add the version onto it.

bika/lims/content/instrument.py
```diff
@@ -55,6 +57,14 @@
 from bika.lims import bikaMessageFactory as _
 from bika.lims import deprecated
 
+def cache_key(method, self):
```
mikejmets: Use api.get_cache_key instead.

bika/lims/content/instrument.py
```diff
@@ -13,6 +13,8 @@
 from Products.CMFPlone.utils import safe_unicode
 from Products.Archetypes.atapi import DisplayList, PicklistWidget
 from Products.Archetypes.atapi import registerType
+from bika.lims.api import bika_cache_key_decorator
+from plone.memoize.volatile import store_on_context, cache
```
ramonski: Please check for unused imports

bika/lims/content/instrument.py
```diff
@@ -55,6 +57,14 @@
 from bika.lims import bikaMessageFactory as _
 from bika.lims import deprecated
 
+def cache_key(method, self):
+    creation_flag = self.checkCreationFlag()
+    if creation_flag:
+        raise DontCache
```
ramonski: I think we missed that in the "official" `bika.lims.api` version. I think we have to add that snippet in this decorater: https://github.com/bikalims/bika.lims/blob/master/bika/lims/api.py#L1060

bika/lims/browser/fields/historyawarereferencefield.py
```diff
@@ -19,9 +19,20 @@
 from Products.CMFEditions.Permissions import AccessPreviousVersions
 from Products.Archetypes.config import REFERENCE_CATALOG
 from bika.lims import bikaMessageFactory as _
+from bika.lims.api import bika_cache_key_decorator, get_uid
 from bika.lims.utils import t
 from bika.lims import logger
 from bika.lims.utils import to_utf8
+from plone.memoize.volatile import store_on_context, cache, DontCache
```
ramonski: better use the `bika.lims.api` like @mikejmets suggested already. Also `store_on_context` is never userd

bika/lims/content/instrument.py
```diff
@@ -481,6 +491,7 @@ def getValidCertifications(self):
                 certs.append(c)
         return certs
 
+    @cache(cache_key)
```
ramonski: Please also note that without the second parameter, the `@cache` decorator stores the value on the calling view. IMO we should use the context instead.

bika/lims/content/instrument.py
```diff
@@ -491,6 +502,7 @@ def isValid(self):
             and self.isValidationInProgress() is False \
             and self.isCalibrationInProgress() is False
 
+    @cache(cache_key)
     def getLatestReferenceAnalyses(self):
```
ramonski: Never cache persistent objects:
https://stackoverflow.com/questions/38794053/what-are-the-possible-causes-of-a-connectionstateerror/38794725#38794725

PR #2279, opened 2017-10-20
ISSUE-2278: Verify calculated results the first time around

Lunga001:

Linked issue: https://github.com/bikalims/bika.lims/issues/2278

## Current behavior before PR
    An AR is verifiable but it has analyses that are on to_be_verified state and when its verified it does 
    not verify all its children(analyses)
    Calculated results that are in to_be_verified state do not verify the first time around
     Calculated results that are in to_be_verified state should verify the first time around
--

### DIFF
--- a/bika/lims/browser/bika_listing.py
+++ b/bika/lims/browser/bika_listing.py
@@ -225,6 +225,16 @@ class WorkflowAction:
         dest = None
         transitioned = []
         workflow = getToolByName(self.context, 'portal_workflow')
+        # NOTE:Calculated result are on the top of the list and because of that
+        # they cannot be transitioned from to_be_verified to verified
+        # because their dependencies have not been transition
+        # so we transition the dependencies first by putting the
+        # calculated results at the end of the list
+        items = sorted(items,
+                key=lambda x: (
+                    hasattr(x, 'calculateResult')
+                    and x.calculateResult(override=True, cascade=True)
+                    is True, x))
 
         # transition selected items from the bika_listing/Table.
         for item in items:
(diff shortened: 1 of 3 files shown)


andersonsmith:
Does this have any effect on issue 1978?

https://github.com/bikalims/bika.lims/issues/1978

Lunga001:
Not sure, but it does seem to have some similarities, this is about the verifying the AR and not editing or submitting the results. Once the analyses are in to_be_verified state they cannot be edited.

bika/lims/content/analysisrequest.py
```diff
@@ -2682,29 +2682,33 @@ def isVerifiable(self):
...
+        for a in self.getAnalyses(full_objects=True):
+            st = workflow.getInfoFor(a, 'cancellation_state', 'active')
+            if st == 'cancelled':
+                continue
+            st = workflow.getInfoFor(a, 'review_state')
+            if st in omit:
+                continue
+            # Can the analysis be verified?
+            if not a.isVerifiable():
+                canbeverified = False
```
ramonski: Why are all the other checks needed if only `a.isVerifiable()` toggles the `canbeverified` switch to `False`?

Also please use the `api` functions, e.g. https://github.com/bikalims/bika.lims/blob/master/bika/lims/api.py#L720

bika/lims/browser/bika_listing.py
```diff
@@ -225,6 +225,16 @@ def submitTransition(self, action, came_from, items):
...
+        # NOTE:Calculated result are on the top of the list and because of that
+        # they cannot be transitioned from to_be_verified to verified
+        # because their dependencies have not been transition
+        # so we transition the dependencies first by putting the
+        # calculated results at the end of the list
+        items = sorted(items,
+                key=lambda x: (
+                    hasattr(x, 'calculateResult')
+                    and x.calculateResult(override=True, cascade=True)
+                    is True, x))
```
ramonski: This feels to me like wrong approach to solve the issue.
Please note that the sorting should be consistent in the whole system. If we activate that code, the "Sort Key" settings of Analyses Services are ignored.

PR #2295, opened 2017-11-08
Normalise ID server results

mikejmets:
## Current behavior before PR
* Unnormalised IDs would break the rename of new objects
* Keys in the number generator are not prefixed with anything which can lead to duplicate keys for different portal types

* IDs are normalised so new objects are created with not illegal characters.
* Prefix number generator keys with the lowercase of the portal type.

--

### DIFF
--- a/bika/lims/content/bikasetup.py
+++ b/bika/lims/content/bikasetup.py
@@ -774,27 +774,7 @@ schema = BikaFolderSchema.copy() + Schema((
         widget=RecordsWidget(
             label=_("Formatting Configuration"),
             allowDelete=True,
-            description=_("""The ID Server in Bika LIMS provides IDs for content items base of the given format specification.
-
-The format string is constructed in the same way as a python format() method
-based predefined variables per content type. The only variable available to all
-type is 'seq'. Currently, 'seq' can be constructed either using number generator
-or a counter of existing items. For generated IDs, one can specify the point at
-which the format string will be split to create the generator key. For counter
-IDs, one must specify the context and the type of counter which is either the
-number of backreferences or the number of contained objects.
-
-Configuration Settings:
-* format:
-  - a python format string constructed from predefined variables like sampleId, client, sampleType.
-  - special variable 'seq' must be positioned last in the format string
-* sequence type: [generated|counter]
-* context: if type counter, provides context the counting function
-* counter type: [backreference|contained]
-* counter reference: a parameter to the counting function
-* prefix: default prefix if none provided in format string
-* split length: the number of parts to be included in the prefix
-""")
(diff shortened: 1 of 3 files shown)


bika/lims/idserver.py
```diff
@@ -133,18 +138,27 @@ def getConfigByPortalType(config_map, portal_type):
...
-        elif config.get('split_length', None) > 0:
-            prefix_config = '-'.join(form.split('-')[:config['split_length']])
-            prefix = prefix_config.format(**variables_map)
-        else:
-            prefix = config['prefix']
-        new_seq = number_generator(key=prefix)
+        try:
+            if config.get('split_length', None) == 0:
+                prefix_config = '{}-{}'.format(portal_type.lower(),
+                                               '-'.join(form.split('-')[:-1]),
```
ramonski: This splitting task seems to be needed everywhere in this code. Please factor that out into a function which also takes care about the `KeyError`, e.g.
```python
def split_and_slice(string, separator="-", start=0, end=None):
    if not isinstance(string, basestring):
        return None
    try:
        segments = string.split(separator)
        if end is None:
            end = len(segments)
        return segments[start:end] 
    except KeyError:
        return None

File: bika/lims/content/bikasetup.py [hunk 2]
```diff
@@ -774,27 +774,7 @@ def getCounterTypes(self, instance=None):
...
...e.g. '03d' for a sequence of IDs from 001 to 999.</p> <p>Alphanumeric prefixes for IDs are included as is in the formats, e.g. WS for Worksheet in WS-{seq:03d} produces sequential Worksheet IDs: WS-001, WS-002, WS-003 etc.</p> <p></p> <p>Variables that can be used include:<table><tr><th style='width:150px'>Content Type</th><th>Variables</th></tr> <tr><td>Client</td><td>{client}</td></tr> <tr><td>Year</td><td>{year}</td></tr> <tr><td>Sample ID</td><td>{sampleId}</td></tr> <tr><td>Sample Type</td><td>{sampleType}</td></tr> <tr><td>Sample Date</td><td>{sampleDate}</td></tr></table></p><p>Configuration Settings:<ul><li>format: <ul><li>a python format string constructed from predefined variables like sampleId, client, sampleType.</li><li>special variable 'seq' must be positioned last in the format string</li></ul></li><li>sequence type: [generated|counter]</li><li>context: if type counter, provides context the counting function</li><li>counter type: [backreference|contained]</li><li>counter reference: a parameter to the counting function</li><li>prefix: default prefix if none provided in format string</li><li>split length: the number of parts to be included in the prefix</li></ul></p>")
```
ramonski: This change is also in here:
https://github.com/bikalims/bika.lims/pull/2297/files/0c7a6d60b992a64bcce7019c7bff23245cc9cc76#diff-0dde5125b635d6fb55f5372798dae571R777

Maybe you can put that in a separate PR

ramonski:
I see some similarities to https://github.com/bikalims/bika.lims/pull/2297.
Do we still need to merge that PR @mikejmets?

ramonski:
I think this is now superseded by https://github.com/bikalims/bika.lims/pull/2311 and https://github.com/bikalims/bika.lims/pull/2312

PR #2296, opened 2017-11-08
AN-155 specifications on off

Lunga001:

https://jira.bikalabs.com/browse/AN-155

## Current behavior before PR
   Dormant Specifications are used. Even if Specifications are all off
    When Enable Analysis Specifications is disabled/off
        Hide the Specification column from Analyses table on AR View 
        On the AR add form and on the edit view, hide the Analysis Specification field

    When Enable Analysis Specifications is enabled/on
        Don't use inactive/dormant specifications
        Show the Specification column from Analyses table on AR View
        On the AR add form user Default Analysis Specifications that are choosen on the Bika Setup Tab,
        like Sample Type Specifications(Lab) for example
     
--

### DIFF
--- a/bika/lims/browser/analyses.py
+++ b/bika/lims/browser/analyses.py
@@ -123,6 +123,9 @@ class AnalysesView(BikaListingView):
                 'index': 'getDueDate',
                 'sortable': False},
         }
+        if not self.is_ar_specs_allowed():
+            if 'Specification' in self.columns:
+                del self.columns['Specification']
 
         self.review_states = [
             {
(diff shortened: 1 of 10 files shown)


Lunga001:
Also please note that I didn't know any better how to show/hide a field on the form  based on a condition. In this case hide/show the Specification field on the AR add form depending on the value of  Enable Analysis Specifications in Bika Set - Analyses Tab.
So instead this is  the solution I had on https://github.com/Lunga001/bika.lims/blob/AN-155-Specifications-on-off/bika/lims/browser/analysisrequest/add2.py#L476 and https://github.com/Lunga001/bika.lims/blob/AN-155-Specifications-on-off/bika/lims/browser/header_table.py#L143

PR #2297, opened 2017-11-08
Added seeding function to IDServer

mikejmets:
## Current behavior before PR
The is no way to change and seed a ID prefex

Seed function that allows a manager to set a seed for a prefix

--


lemoene:
!!

bika/lims/browser/idserver/view.py
```diff
@@ -0,0 +1,22 @@
...
+        """
+        form = self.request.form
+        prefix = form.get('prefix', None)
+        if prefix is None:
+            return 'No prefix provided'
+        seed = form.get('seed', None)
+        if seed is None:
+            return 'No seed provided'
+
+        seed = int(seed) - 1
```
ramonski: Please be more defensive at that point before you do the type conversion, e.g. do a check for `seed.isdigit()` and afterwards a check if the number is `>0`

bika/lims/docs/IDServer.rst
```diff
@@ -131,7 +131,7 @@ Set up `ID Server` configuration::
     ...             'portal_type': 'SamplePartition',
     ...             'sequence_type': 'counter',
     ...             'value': ''},
-    ...            {'form': 'B{year}-{seq:04d}',
+    ...            {'form': 'BÖ-{year}-{seq:04d}',
```
ramonski: I think that's an artifact from testing?

bika/lims/numbergenerator.py
```diff
@@ -91,6 +91,31 @@ def get_number(self, key):
         logger.debug("NUMBER after => %s" % storage.get(key, '-'))
         return storage[key]
 
+    def set_seed(self, key, seed=None):
+        """ get the next consecutive number
+        """
+        storage = self.storage
+
+        if seed:
```
ramonski: reduce code complexity and return immediately here.

bika/lims/numbergenerator.py
```diff
@@ -91,6 +91,31 @@ def get_number(self, key):
...
+                    logger.debug("NUMBER => %d" % storage[key])
+            finally:
+                logger.debug("*** consecutive number lock release ***")
+                self.storage._p_changed = True
+                lock.release()
+
+            logger.debug("In set_seed NUMBER => %d" % storage[key])
+        return storage[key]
+
+
```
ramonski: This whole function duplicates the code from `get_number` except of one line, which is not so nice. Maybe a dedicated setter could be better?

bika/lims/browser/idserver/view.py
```diff
@@ -0,0 +1,22 @@
+from bika.lims.browser import BrowserView
+from bika.lims.numbergenerator import INumberGenerator
+from zope.component import getUtility
+
+
+class IDServerView(BrowserView):
+
```
ramonski: Please add a proper docstring with usage example

bika/lims/content/bikasetup.py
```diff
@@ -774,27 +774,7 @@ def getCounterTypes(self, instance=None):
...
...e.g. '03d' for a sequence of IDs from 001 to 999.</p> <p>Alphanumeric prefixes for IDs are included as is in the formats, e.g. WS for Worksheet in WS-{seq:03d} produces sequential Worksheet IDs: WS-001, WS-002, WS-003 etc.</p> <p></p> <p>Variables that can be used include:<table><tr><th style='width:150px'>Content Type</th><th>Variables</th></tr> <tr><td>Client</td><td>{client}</td></tr> <tr><td>Year</td><td>{year}</td></tr> <tr><td>Sample ID</td><td>{sampleId}</td></tr> <tr><td>Sample Type</td><td>{sampleType}</td></tr> <tr><td>Sample Date</td><td>{sampleDate}</td></tr></table></p><p>Configuration Settings:<ul><li>format: <ul><li>a python format string constructed from predefined variables like sampleId, client, sampleType.</li><li>special variable 'seq' must be positioned last in the format string</li></ul></li><li>sequence type: [generated|counter]</li><li>context: if type counter, provides context the counting function</li><li>counter type: [backreference|contained]</li><li>counter reference: a parameter to the counting function</li><li>prefix: default prefix if none provided in format string</li><li>split length: the number of parts to be included in the prefix</li></ul></p>")
```
ramonski: Please split that line into multiple lines, so that it is readable in the editor as well.
Hint: You don't need backslashes in between parenthesis.

bika/lims/docs/IDServer.rst
mikejmets: It is an artefact from testing but I thought it was worth leaving in because it broadens the test.

bika/lims/content/bikasetup.py
mikejmets: Splitting into multiple lines using triple quotes doesn't work in a description. I don't get your hint, please explain.

bika/lims/browser/idserver/view.py
mikejmets: I've added isdigit but I'm not sure if we should restrict the seed to start as a positive number. Are you sure?

bika/lims/content/bikasetup.py
ramonski: Hi Mike, I meant something like this:

```python
 description=_("<p>The Bika LIMS ID Server provides unique sequential IDs for objects "
                         "such as Samples and Worksheets etc, based on a format specified for "
                         "each content type.</p> <p>The format is constructed similarly to the " 
                         " .... </p>")
```

bika/lims/browser/idserver/view.py
ramonski: Yes, I think negative numbers don't make sense and I would feel better if we do not allow this level of flexibility;)

bika/lims/numbergenerator.py
mikejmets: Done

bika/lims/numbergenerator.py
mikejmets: Done

bika/lims/browser/idserver/view.py
mikejmets: Done

bika/lims/browser/idserver/view.py
mikejmets: Done

bika/lims/content/bikasetup.py
mikejmets: Done

bika/lims/browser/idserver/view.py
```diff
@@ -0,0 +1,31 @@
...
+        prefix = form.get('prefix', None)
+        if prefix is None:
+            return 'No prefix provided'
+        seed = form.get('seed', None)
+        if seed is None:
+            return 'No seed provided'
+        if not seed.isdigit():
+            return 'Seed must be a digit'
+        if not seed.isdigit():
+            return 'Seed must be a digit'
```
ramonski: I think one check is enough;)

bika/lims/browser/idserver/view.py
```diff
@@ -0,0 +1,31 @@
...
+        seed = int(seed)
+        if seed < 0:
+            return 'Seed cannot be negative'
+        seed = seed - 1
```
ramonski: Take care if the user adds `seed=0`

bika/lims/numbergenerator.py
```diff
@@ -91,6 +91,23 @@ def get_number(self, key):
...
+        """ set a key's value
+        """
+        storage = self.storage
+
+        if not isinstance(value, int):
+            logger.error("set_number: Value must be an integer")
+            return
+
+        lock.acquire()
+        storage[key] = value
```
ramonski: Take care of dead-locks if you get an error here. Maybe use `try..finally` or check for unhashable keys

bika/lims/idserver.py
```diff
@@ -82,6 +75,20 @@ def getConfigByPortalType(config_map, portal_type):
...
         return config
 
+    def splitSliceJoin(string, separator="-", start=0, end=None):
+        if not isinstance(string, basestring):
+            return None
+        try:
+            segments = string.split(separator)
+            if end is None:
+                end = len(segments)
+            return separator.join(segments[start:end])
```
ramonski: not sure if relevant, but maybe take care that `start<end`, `start>=0`, `end<len(segments)` etc.

bika/lims/idserver.py
```diff
@@ -5,21 +5,26 @@
...
+import zLOG
 from DateTime import DateTime
 from Products.ATContentTypes.utils import DT2dt
+from OFS.CopySupport import CopyError
 
 from bika.lims import api
-# from bika.lims import logger
 from bika.lims import bikaMessageFactory as _
+from bika.lims import logger
+from bika.lims.browser.fields.uidreferencefield import get_backreferences
```
ramonski: ups, this one has to go away for `bikalims/bika.lims`, because that is only available in `senaite/bika.lims`

bika/lims/browser/idserver/view.py
mikejmets: If seed = 0 then next ID will be 0. I would then there are cases where starting from 0 is required.

bika/lims/idserver.py
mikejmets: Removed backreferences but had to leave logger because it's used.

PR #2298, opened 2017-11-09
AN-145 initialise batch title field empty

Lunga001:
https://jira.bikalabs.com/browse/AN-145

## Current behavior before PR
     On the batchfolder(batch listing) the Batch Title is empty
     If Batch Title is not filled in it should have the same value as Batch ID

--

### DIFF
--- a/bika/lims/content/batch.py
+++ b/bika/lims/content/batch.py
@@ -222,9 +222,12 @@ class Batch(ATFolder):
     def Title(self):
         """ Return the Batch ID if title is not defined """
         titlefield = self.Schema().getField('title')
-        if titlefield.widget.visible:
+        titlevalue = titlefield.get(self)
+        if titlefield.widget.visible and titlevalue:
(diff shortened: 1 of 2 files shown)


bika/lims/content/batch.py
```diff
@@ -222,9 +222,12 @@ def _renameAfterCreation(self, check_auto_id=False):
     def Title(self):
         """ Return the Batch ID if title is not defined """
         titlefield = self.Schema().getField('title')
-        if titlefield.widget.visible:
+        titlevalue = titlefield.get(self)
+        if titlefield.widget.visible and titlevalue:
             return safe_unicode(self.title).encode('utf-8')
         else:
+            if self.id.startswith('batch.'):
+                return ''
```
ramonski: Please use this code instead:

```
pu = api.portal.get_tool("plone_utils")
pu.isIDAutoGenerated(self.getId())
```

Lunga001: Thank you @ramonski

PR #2306, opened 2017-11-14
do not use localized date for chart js.

veroc:
Do not use localized dates for control chart as it breaks the controlchart.js datetime parser.

bika.lims.graphics.controlchart.js
// Convert values to floats
// "2014-02-19 03:11 PM"
x_data_parse = d3.time.format("%Y-%m-%d %I:%M %p").parse;

### DIFF
--- a/bika/lims/browser/referencesample.py
+++ b/bika/lims/browser/referencesample.py
@@ -167,6 +167,7 @@ class ReferenceAnalysesView(AnalysesView):
         # they are not actually used in the table rendering.
         item['Keyword'] = service.getKeyword()
         item['Unit'] = service.getUnit()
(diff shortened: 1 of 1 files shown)


bika/lims/browser/referencesample.py
```diff
@@ -167,6 +167,7 @@ def folderitem(self, obj, item, index):
         # they are not actually used in the table rendering.
         item['Keyword'] = service.getKeyword()
         item['Unit'] = service.getUnit()
+        item['CapturedRaw'] = obj.getResultCaptureDate().strftime('%Y-%m-%d %I:%M %p')
```
ramonski: Hi @veroc, did you test this also when the result has no capture date?

veroc: I assumed that there're only results with capture date in this context. Now it returns an empty string on missing date.

PR #2315, opened 2017-11-20
IDServer doesn't find next correct sequence after flush

mikejmets:
The ID Server does not find the next correct sequence after flushing the number generator

Linked issue: https://github.com/bikalims/bika.lims/issues/2314

## Current behavior before PR
For a ID Template that has multiple segments, after a flush, the wrong sequence is returned

Calculate the next sequence number correctly.
--


bika/lims/idserver.py
```diff
@@ -249,11 +263,12 @@ def get_counted_number(context, config, variables, **kw):
     number = len(seq_items)
     return number
 
-
 def get_generated_number(context, config, variables, **kw):
     """Generate a new persistent number with the number generator for the
     sequence type "Generated"
     """
+    separator = '-'
```
ramonski: Make this one `separator=kw.get('separator', '-')`

bika/lims/idserver.py
```diff
@@ -225,6 +224,21 @@ def get_ids_with_prefix(portal_type, prefix):
...
+    segments = split(id_template, separator)
+    seq_index = 0
+    for idx in range(len(segments)):
+        segment = segments[idx]
+        segment = segment.replace('{', '')
+        segment = segment.replace('}', '')
+        if segment.split(':')[0] == 'seq':
+            seq_index = idx
+            break
+    return seq_index
```
ramonski: I would rather go for a regex search per segment, but ok for now

PR #2323, opened 2017-12-04
NG flush breaks id server

mikejmets:
Items in the bika_setup_catalog do not continue in sequence after flushing the number generator

Linked issue: https://github.com/bikalims/bika.lims/issues/2322

## Current behavior before PR
After a flush, BSC items get new IDs from 1 instead of the next in the sequence

All items continue in the sequence after ng_flush

--

### DIFF
--- a/bika/lims/idserver.py
+++ b/bika/lims/idserver.py
@@ -209,11 +209,26 @@ def get_current_year():
     return DateTime().strftime("%Y")[2:]
 
 
+def search_catalogs(portal_type):
+    """Returns brains which share the same portal_type
+    """
+    catalog_names = ['portal_catalog', 'bika_setup_catalog', 'bika_catalog']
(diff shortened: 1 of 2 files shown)


bika/lims/idserver.py
```diff
@@ -209,11 +209,26 @@ def get_current_year():
     return DateTime().strftime("%Y")[2:]
 
 
+def search_catalogs(portal_type):
```
ramonski: Please use this function instead: https://github.com/bikalims/bika.lims/blob/master/bika/lims/api.py#L582

mikejmets: At first glance it looks like api.search allows a list of catalog names but then fails if more than one is given. Should I extend api.search to all multiple catalogs or simply change idserver to make multiple calls to api.search?

